PR #28, opened 2020-08-03
Solution

krajnova:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const reversed = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (reversed[value] !== undefined) {
+      return null;
```
DmJSCourse, line 33: It works, but you can also try
https://developer.mozilla.org/uk/docs/Web/JavaScript/Reference/Operators/in
for a cleaner solution

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+    } else {
```
DmJSCourse, line 34: else is not necessary here, but it does not really matter

src/inverseRobot.js
krajnova: Oh, thx. This is really useful)

src/inverseRobot.js
krajnova: I fixed it. Thanks)

PR #31, opened 2020-08-03
solution

furude-rika:


src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+  const arr = Object.values(robot);
+  let firstElem = arr[0];
+
+  for (let i = 1; i < arr.length; i++) {
+    if (firstElem !== arr[i]) {
+      firstElem = arr[i];
+    } else {
+      return null;
+    }
+  }
```
vsm2k17, line 37: you can repalce this with newObj.hasOwnProperty(robot[key]) inside of loop below

PR #35, opened 2020-08-03
Solution

asbh25:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const [key, value] of Object.entries(robot)) {
```
vkryvytskyy, line 29: nice move, but maybe `robot[key]` is not so ugly to use Object.entries

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+    if (value in inversed) {
+      return null;
```
vkryvytskyy, line 31: use `continue` keyword

PR #39, opened 2020-08-03
Solution

rmnkk:


src/inverseRobot.js
```diff
@@ -25,6 +25,24 @@
  */
 function inverseRobot(robot) {
   // write code here
+  let reversed = {};
+  const arrayOfProp = [];
+
+  for (const prop in robot) {
+    arrayOfProp.push(robot[prop]);
+
+    if (robot.hasOwnProperty(prop)) {
```
vsm2k17, line 34: prop that's a property which you just pull out of the robot, what exactly you want to get form this checking?

src/inverseRobot.js
```diff
@@ -25,6 +25,24 @@
...
+      reversed[robot[prop]] = prop;
+    }
+  }
+
+  for (const elem of arrayOfProp) {
+    if (arrayOfProp.indexOf(elem) !== arrayOfProp.lastIndexOf(elem)) {
+      reversed = null;
+    }
+  }
```
vsm2k17, line 43: read about hasOwnProperty

src/inverseRobot.js
rmnkk: Got it. I thought that I need to check if the object has the property which I'm going to change.  
But I loop through these properties before, so they are in the object. 
Then this condition is redundant.

src/inverseRobot.js
rmnkk: Got it. I have used hasOwnProperty where it wasn't actually necessary.

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
  */
 function inverseRobot(robot) {
   // write code here
+  let reversed = {};
+  const arrayOfProp = [];
+
+  for (const prop in robot) {
+    arrayOfProp.push(robot[prop]);
```
vsm2k17, line 32: you can make your code prettier if you check here: got your object this property or not, and if it is got this property than you can immediately return null. This action  let you avoid the second loop

PR #41, opened 2020-08-03
Solution

aliceblunt8:


src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
-  // write code here
+  const twin = {};
+
+  // Iterate through robot's keys and copy all props to twin
+  for (const key in robot) {
+    twin[robot[key]] = key;
+  }
+
+  const allKeys = Object.keys(robot);
+  const uniqueKeys = Object.keys(twin);
```
vsm2k17, line 35: you can just check twin.hasOwnProprty(robot[key]) in the loop above before you saved reversed values into a twin.

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+
+  // Check for duplicate keys and return null if found
+  if (allKeys.length !== uniqueKeys.length) {
```
vsm2k17, line 38: redundant

src/inverseRobot.js
aliceblunt8: thanks for the tip, now code looks neat!

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const twin = {};
+
+  for (const key in robot) {
+    if (twin.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      twin[robot[key]] = key;
+    }
```
vsm2k17, line 34:
```suggestion
    }
    twin[robot[key]] = key;
```

vsm2k17: no need to write "else" after return

PR #44, opened 2020-08-03
Solution

kostyasabada:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]] === undefined) {
```
vkryvytskyy, line 30:
```suggestion
    if (!newRobot[robot[key]]) {
```

vkryvytskyy: if you don't expect that newRobot has key with null value

PR #46, opened 2020-08-03
Solution

oivannikov:


src/inverseRobot.js
```diff
@@ -23,8 +23,23 @@
...
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+  const values = Object.values(robot);
+  const keys = Object.keys(robot);
+
+  values.reverse();
+  keys.reverse();
```
vsm2k17, line 33: Screenshot 2020-08-05 at 14 37 30
no need to change the order of fields inside an object. you need to take a key and swap  it with value, they in the same order as they were before

src/inverseRobot.js
```diff
@@ -23,8 +23,23 @@
...
+
+  for (let i = 0; i < values.length; i++) {
```
vsm2k17, line 35: try to use for in, that's let you avoid variables: values and keys, no need to use Object.values(robot);

src/inverseRobot.js
```diff
@@ -23,8 +23,23 @@
...
+    if (newObj.hasOwnProperty(values[i])) {
+      return null;
+    }
+    newObj[values[i]] = keys[i];
```
vsm2k17, line 39: if you write "for (const key in robot)"
```suggestion
    newObj[robot[key]] = key;
```

src/inverseRobot.js
```diff
@@ -23,8 +23,23 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
```
vsm2k17, line 28: I am sure you can name your variables better.

PR #52, opened 2020-08-03
Solution

martakupka:


src/inverseRobot.js
```diff
@@ -24,7 +24,32 @@
...
+  const valuesArr = [];
+  let count = 0;
+  let i = 0;
+  let j = 0;
+
+  for (const key in robot) {
+    keysArr.push(key);
+    valuesArr.push(robot[key]);
+    count++;
+  }
```
vsm2k17, line 38: you can do everything inside this loop
instead of pushing key and value somewhere else, you can assign them into new object
and before assignment check, if your new object has the property that you want to assign, if it is, then return null

src/inverseRobot.js
```diff
@@ -24,7 +24,32 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const keysArr = [];
+  const valuesArr = [];
+  let count = 0;
+  let i = 0;
+  let j = 0;
```
vsm2k17, line 32: redundant

PR #56, opened 2020-08-03
inverse

yuriimovchko:


src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const res = {};
+
+  for (const i in robot) {
+    if (res[robot[i]] !== undefined) {
```
vsm2k17, line 31:
```suggestion
    if (res[robot[i]]) {
```

src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const res = {};
+
+  for (const i in robot) {
```
vsm2k17, line 30: Use more meaningful names, "i" not a good name, something like "key" would be fine

PR #66, opened 2020-08-04
Solution

mpone:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+  const valuesInSort = values.sort();
```
vkryvytskyy, line 28: you don't need to use sorting in this task, use nested loops

PR #72, opened 2020-08-06
solution

Paavlo:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+
+  for (const key in robot) {
+    reversRobot[robot[key]] = key;
+  }
+
+  if (Object.entries(reversRobot).length !== values.length) {
+    return null;
+  } else {
+    return reversRobot;
+  }
```
vkryvytskyy, line 38: use @Elena-brv suggestion and replace this by hasOwnProperty

PR #158, opened 2020-12-04
Solution

rostyslav-horytskyi:


src/inverseRobot.js
```diff
@@ -25,6 +25,31 @@
...
+
+  for (const item of arr) {
+    if (!result.includes(item)) {
+      result.push(item);
+    }
+  }
+
+  if (result.length !== arr.length) {
+    return null;
+  }
```
vkryvytskyy, line 50: if(!res[robot[key]]) {
  return null;
} else {
  res[robot[key]] = key;
}

vkryvytskyy (changes requested):

src/inverseRobot.js
rostyslav-horytskyi: > if(!res[robot[key]]) {
> return null;
> } else {
> res[robot[key]] = key;
> }


I wrote like this:

const res = {};

  for (const key in robot) {
    if (!res[robot[key]]) {
      return null;
    } else {
      res[robot[key]] = key;
    }
  }

return res;

Not all test pass((

rostyslav-horytskyi:
I realized what I was doing wrong)

src/inverseRobot.js
```diff
@@ -25,6 +25,18 @@
...
+
+  const res = {};
+
+  for (const key in robot) {
+    if (res[robot[key]]) {
+      return null;
+    } else {
+      res[robot[key]] = key;
+    }
+  }
```
vpolets, line 37:
```suggestion
    } 
      res[robot[key]] = key;
  }
```

vpolets (approved):

PR #166, opened 2020-12-08
solution

dtsehelnyk:


kerpichonka (approved):

src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
```
varseniuk, line 32: Try to make your function more simple. Now it copies EVERY value of the robot, even if it's repeated. And after that, you compare both objects' lengths.
You may iterate through the object's keys with `for...in...` loop, check if newRobot already has a current property (using that key) and if NOT - add value to the new robot. If the value already present - return null.

Your solution works and it's ok but try to make it a little more laconic

varseniuk (changes requested):
It's ok, but try to make it simpler

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
```
varseniuk, line 27:
```suggestion

```
remove all commented lines before pushing solutions

varseniuk (approved):

PR #167, opened 2020-12-08
Solution

troospieler:


MarchD (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let i = 0;
```
varseniuk, line 27: let's use more meaningful variables names.
If it is the count of robot parts, name it smth like 'counter'. Name 'i' mean's nothing.

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+  const newRobot = {};
+
+  for (const each in robot) {
```
varseniuk, line 30: The same. Name 'each' not meaningful. Chose a more suitable name for this variable (e.g. 'part', 'value' etc...)

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+    newRobot[robot[each]] = each;
+    i++;
+  }
+
+  const a = Object.keys(newRobot).length;
+
+  if (a !== i) {
+    return null;
+  }
```
varseniuk, line 39:
```suggestion
  if (i !== Object.keys(newRobot).length) {
    return null;
  }
```

There's no need to create a new variable in such single case.

varseniuk (changes requested):

varseniuk (approved):

PR #175, opened 2020-12-10
solution

xipholena:


5Mountains (approved):
Awesome solution!

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObject = {};
```
varseniuk, line 27: try to give variables more meaningful names. E.g. in this case 'newRobot' or something like that

varseniuk (approved):
Good 👍

PR #178, opened 2020-12-10
Inversed Robots

tamara-ostapets:


paramonovfrontend (approved):

src/inverseRobot.js
```diff
@@ -24,7 +25,18 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const inRobot = {};
+
+  const entries = Object.entries(robot);
+
+  for (let i = 0; i < entries.length; i++) {
+    if (inRobot.hasOwnProperty(entries[i][1])) {
+      return null;
```
varseniuk, line 34: You could make it more simple:
`  for (const entries in robot) {
    if (newRobot.hasOwnProperty(robot[entries])) {
      return null;`

And there would be no need in 'Object.entries(robot)'

PR #181, opened 2020-12-10
Solution

vitalya2626:


zhenya-mezhueva (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
+
+  for (const x in robot) {
```
varseniuk, line 29:
```suggestion
  const copiedRobot = {};

  for (const part in robot) {
```

Try to give more meaningful names to variables

varseniuk (approved):
All is ok, but in the future try to give variables names that mean their purposes

src/inverseRobot.js
vitalya2626: Okay, thanks!

PR #188, opened 2020-12-10
 Solution InverseRobot

Nvalja:


zhenya-mezhueva (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const repairedRobot = {};
+
+  for (const key in robot) {
+    if (repairedRobot.hasOwnProperty([robot[key]])) {
+      return null;
+    } else {
+      repairedRobot[robot[key]] = key;
+    }
```
varseniuk, line 34:
```suggestion
    }
    repairedRobot[robot[key]] = key;
```

No need for this `else` because if the first check passes, the function will return 'null' and stop, after that no further conditions won't execute

varseniuk (changes requested):

varseniuk (approved):
👍

PR #189, opened 2020-12-10
Solution

AlieksieienkoAndriy:


SofiiaAsieieva (approved):
:+1:

src/inverseRobot.js
```diff
@@ -23,8 +23,25 @@
...
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const robotData = Object.entries(robot);
+
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < (robotValues.length - 1); i++) {
+    if (robotValues[i] === robotValues[i + 1]) {
```
vsm2k17, line 33: you just return null if the value the same as the next one, lets try to solve this without robotData and robotValue arrays and using only one loop.
*hint: you need to use that kind of loop which can give access to the object key

vsm2k17 (changes requested):

src/inverseRobot.js
AlieksieienkoAndriy: Thanks for the good advice!

vsm2k17 (approved):
good job

PR #194, opened 2020-12-10
Solution

taniazt:


olesiatk (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+  const set = new Set(values);
+
+  if (values.length !== set.size) {
+    return null;
+  }
```
varseniuk, line 32: It's good that you already know how to use Set construction 👍 
But let's make this simple task without it and using in it more simple operations.

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+
+  const invRobot = {};
+
+  Object.keys(robot).forEach(key => {
+    const newKey = robot[key];
+
+    invRobot[newKey] = key;
+  });
```
varseniuk, line 40: Here you can just check if the robot already has the same own property and if it does - return 'null', 
in other case - `invRobot[newKey] = key;` as you've already wrote.

And it's way more simple that use Set and Object.values and compare their sizes.

varseniuk (changes requested):

Denchepornyuk (approved):

PR #196, opened 2020-12-11
solution

Pasha5555:


sasha39612 (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+  const revRobot = {};
+
+  for (let i = 0; i < robotValues.length; i++) {
+    if (robotValues[i] === robotValues[i + 1]) {
+      return null;
+    }
+  }
```
varseniuk, line 34: You can make this part shorter and simple just using `for...in...` loop and check if the new object already has the iterable property. And if so you will not need a variable `robotValues`

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  for (const key in robot) {
+    revRobot[robot[key]] = key;
+  }
```
varseniuk, line 38: 👍

varseniuk (changes requested):

src/inverseRobot.js
Pasha5555: Виправив, але в таблиці r автоматично не з'явилося

varseniuk (approved):

PR #200, opened 2020-12-11
Solution

Baburineu:


src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const p in robot) {
```
vkryvytskyy, line 30:
```suggestion
  for (const property in robot) {
```
what does p means?

vkryvytskyy (approved):

src/inverseRobot.js
Baburineu: p means property. Next time I will write more clearly.

PR #210, opened 2020-12-14
Solution

viktor-morhun:


src/inverseRobot.js
```diff
@@ -1,30 +1,22 @@
...
-  // write code here
+  let counter = 1;
+
+  for (const key of Object.keys(robot)) {
+    const arr = Object.values(robot);
+
+    for (let i = counter; i < arr.length; i++) {
+      if (robot[key] === arr[i]) {
+        return null;
+      }
```
andrew-zaycev, line 12: You can use `hasOwnProperty` here, then get rid of `let counter`, and the second `for`

andrew-zaycev (approved):

ShyianO (approved):

aionova (changes requested):
You need to return a new object with all the robots properties reversed.

aionova (approved):
good job!

PR #216, opened 2020-12-24
Solution

AllaByshkina:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const robotKeys = Object.keys(robot);
+  const robotValues = Object.values(robot);
+  const newRobot = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
```
vkryvytskyy, line 31: consider to use for..in loop for objects

vkryvytskyy (approved):

PR #262, opened 2021-02-24
Solution

MaksymMatuliak:


MaksymMatuliak:
I don't know why but I had a problem with style in the test, if it's my fault let me know

vkryvytskyy (approved):

PR #266, opened 2021-03-01
Solution

herkoss:


aionova (changes requested):
valuesArr, valuesSet, resObj, key are bad variable names. Use names that describe what the variable contains, not its type. 
'resObj' doesn't tell me what this is at all. 'inversedRobot' would tell me much more, wouldn't it?

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const arrayOfRobotsValues = Object.values(robot);
```
```suggestion
  const robotPropertiesValues = Object.values(robot);
```

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  const uniqueOfRobotsValues = new Set(arrayOfRobotsValues);
```
aionova, line 29:
```suggestion
  const uniqueRobotPropertiesValues = new Set(arrayOfRobotsValues);
```

aionova (approved):
Great use of Set!

PR #269, opened 2021-03-02
Solution

danylo-onishchuk:


src/inverseRobot.js
```diff
@@ -19,12 +19,26 @@
...
+        return null;
+      }
+    }
+  }
+
+  for (let i = 0; i < robotValues.length; i++) {
+    newRobot[robotValues[i]] = robotKeys[i];
+  }
+
+  return newRobot;
```
danheim, line 41: over complicated
you need only one loop and 1 new object

danheim (changes requested):

danheim (approved):

PR #271, opened 2021-03-02
solution

Serhii-Nikitin:


src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+  let listOfValue = '';
```
danheim, line 29: you don't need it

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+
+  for (const key in robot) {
+    if (!listOfValue.includes(robot[key])) {
```
danheim, line 32: try to check key in the `newRobot` object

danheim (changes requested):

danheim (approved):

PR #274, opened 2021-03-03
Solution

IhorDrul:


src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
...
+  const newRobot = {};
+  const robotKeys = Object.keys(robot);
+
+  for (let i = 0; i < robotKeys.length; i++) {
+    for (let j = i + 1; j < robotKeys.length; j++) {
+      if (robot[robotKeys[i]] === robot[robotKeys[j]]) {
+        return null;
+      }
+    }
+  }
```
danheim, line 37: don't nest loops

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
...
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
```
danheim, line 41: you need only one loop

danheim (changes requested):

danheim (approved):
it could be easier

PR #276, opened 2021-03-03
solution

MaxZadorozhniy:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
```
danheim, line 27: nothing wrong with current robot, just inverted
```suggestion
  const invertedRobot = {};
```

danheim (approved):

PR #277, opened 2021-03-03
'Solution'

Uanov:


src/inverseRobot.js
```diff
@@ -24,7 +24,28 @@
...
+
+  const newRobot = {};
+
+  Object.keys(robot).forEach(function(value) {
+    const key = robot[value];
+
+    newRobot[key] = value;
+  });
+
+  return newRobot;
```
danheim, line 48: over complicated
you need only one loop with one condition

danheim (changes requested):

danheim (approved):

PR #280, opened 2021-03-03
Solution

vladdezhniuk:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let result;
```
danheim, line 27: what is result? array? object? string?

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+  const values = Object.values(robot);
+  const check = new Set(values);
+
+  if (check.size !== values.length) {
+    return null;
+  } else {
```
danheim, line 33: remove else

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+    result = Object.entries(robot);
+
+    for (let a of result) {
+      a = a.reverse();
```
danheim, line 37: nested loop, avoid it

danheim (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let result = [];
+  const values = Object.values(robot);
+  const check = new Set(values);
```
danheim, line 29: what does check mean?
```suggestion
  const uniqueValues = new Set(values);
```

danheim (changes requested):

danheim (approved):
existing simpler solution

PR #281, opened 2021-03-03
solution

Daaiif:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (robot[key] in result) {
+      return null;
+    } else {
```
danheim, line 32: remove else, you don't need it after return

danheim (approved):

PR #284, opened 2021-03-03
Created function inverseRobot

haiduk2019:


src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    const change = robot[key];
```
danheim, line 31:
```suggestion
    const value = robot[key];
```

danheim (approved):
existing simpler solution

PR #287, opened 2021-03-03
done

TarasChaban:


src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  let newRobotValue;
```
danheim, line 28: why did you create it here?

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  for (const [key, value] of Object.entries(robot)) {
+    newRobotValue = robot[key];
```
danheim, line 31: you don't need it

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+    if (newRobot.hasOwnProperty(value)) {
+      return null;
+    } else {
+      newRobot[newRobotValue] = key;
```
danheim, line 36: use value 
```suggestion
      newRobot[value] = key;
```

danheim (changes requested):

danheim (approved):

PR #292, opened 2021-03-09
Solution

Taras-Konovalov:


Burunduk003 (approved):
Interesting solution. goog job!

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
-  // write code here
+  const obj = {};
+  const robotValues = Object.values(robot);
+  const noRepeated = robotValues.filter((item, index) => {
+    return robotValues.indexOf(item) === index;
+  });
+
+  if (robotValues.length !== noRepeated.length) {
+    return null;
+  }
```
aliceblunt8, line 35: it's not a boolean, it's still an array

```suggestion
  const uniqueValues = robotValues.filter((item, index) => {
    return robotValues.indexOf(item) === index;
  });

  if (robotValues.length !== uniqueValues.length) {
    return null;
  }
```

aliceblunt8 (approved):
good job

PR #300, opened 2021-03-13
Solution

vitaliy-buchinchik:


src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const valuesArr = Object.values(robot);
```
aionova, line 27: It's not valuesArr. It's inversedRobot

aionova (changes requested):
Let's do it with just 1 loop.
Also please name your variables properly

aliceblunt8 (approved):

PR #304, opened 2021-03-15
Solution

IvanGrekov:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const objInverse = {};
```
```suggestion
  const inversedRobot = {};
```

aionova (changes requested):
don't use data type in variable names. 
'i' is a bad name for variable too)

aionova (approved):
Good job!

PR #307, opened 2021-03-18
Develop

YuriiLiso:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const trueRobot = {};
+
+  for (const key in robot) {
+    if (trueRobot.hasOwnProperty(robot[key]) === false) {
```
aliceblunt8, line 30: hasOwnProperty returns boolean by default, no need to use `=== false` here

```suggestion
    if (trueRobot.hasOwnProperty(robot[key])) {
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #325, opened 2021-04-21
Solution

igor-stasiv:


Ihor-Karpyn (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const repairedRobot = {};
+
+  for (const key in robot) {
+    const correctKey = robot[key];
+    const correctValue = key;
```
VinnieJ-2k20, line 31: There's no need to introduce extra variable here, simply use `key`

VinnieJ-2k20 (approved):
Good job 👍

PR #328, opened 2021-04-21
first solution

scarlettjohanssons:


src/inverseRobot.js
```diff
@@ -25,6 +25,32 @@
...
+        wordRepeat++;
+      }
+    }
+    count++;
+
+    if (wordRepeat >= 2) {
+      return null;
+    }
+    wordRepeat = 0;
+  }
```
BoikoEva, line 47: try to do the same with loop if and [robot[key]] in your for loop

src/inverseRobot.js
```diff
@@ -25,6 +25,32 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const reverseRobot = {};
+  const robotValues = Object.values(robot);
+  let count = 0;
+  let wordRepeat = 0;
+
+  while (count < robotValues.length) {
```
aliceblunt8, line 33: use just for loop instead of while loop

src/inverseRobot.js
```diff
@@ -25,6 +25,32 @@
...
+    const value = robotValues[count];
+
+    for (let i = 0; i < robotValues.length; i++) {
+      if (value === robotValues[i]) {
+        wordRepeat++;
+      }
+    }
+    count++;
```
aliceblunt8, line 41: let's simplify this, add new values to reverseRobot directly and check for duplicates in ONE for loop

aliceblunt8 (changes requested):

aliceblunt8 (approved):

aionova (approved):

PR #331, opened 2021-04-21
Develop

RuslanaPodshyvalova:


BoikoEva:
Nice!

src/inverseRobot.js
```diff
@@ -23,8 +23,35 @@
...
  */
+
+function checkUniqueVals(arr) {
+  let duplicate = null;
+
+  arr.forEach((item, indx, self) => {
+    if (self.indexOf(item) !== indx) {
+      duplicate = item;
+
+      return 0;
```
VinnieJ-2k20, line 34: `forEach` method doesn't return anything, so it is useless to return values from its callback.

VinnieJ-2k20 (approved):
Good thing that you already use array methods.

I'll approve cause it works correctly.
But for this particular task your solution is too complicated.
You have two separate loops and one of them contains one more loop inside of it.
Try to solve this task using only one plain loop.

When you have spare time you can read about Big O Notation [here](https://www.digitalocean.com/community/tutorials/js-big-o-notation) or go through the algorithms course on Mate platform.

PR #336, opened 2021-04-21
solution

Allrightman:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const duplicates = Object.values(robot);
+  const repairedRobot = {};
+
+  if (duplicates.filter((item, index) =>
```
aionova, line 30: 'some' would fit better here than filter method

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+    duplicates.indexOf(item) !== index).length) {
+    return null;
+  }
+
+  for (const [key, value] of Object.entries(robot)) {
```
aionova, line 35: why not use 'for in' loop here?

aionova (changes requested):
Good start, but let's refactor it to make the code easier to read

src/inverseRobot.js
Allrightman: really, thanks - I learned a new method

src/inverseRobot.js
Allrightman: I tried - it didn't work (due to incorrect writing), but I already figured it out

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const duplicates = Object.values(robot);
```
```suggestion
  const robotProperties = Object.values(robot);
```

aionova (approved):
Good job!

PR #351, opened 2021-04-22
Solution

progmari:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (fixedRobot[robot[key]] !== undefined) {
+      return null;
+    }
+    fixedRobot[robot[key]] = key;
+  }
```
ched88xtc, line 34:
```suggestion
    }
    
    fixedRobot[robot[key]] = key;
  }
```

ched88xtc (approved):

VinnieJ-2k20 (changes requested):
`fixedRobot[robot[key]] !== undefined` is not a reliable check. 
It works in this particulat task, but object can have actual property with undefined value.

Use different check method.

VinnieJ-2k20 (changes requested):
Unfortunately, the check you selected is even less reliable than the previous one.

Hint: use `hasOwnProperty` method.

VinnieJ-2k20 (approved):
Good job 👍

PR #352, opened 2021-04-22
'Solution'

ViPiven:


progmari (approved):
👍

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const res = {};
+
+  for (const key in robot) {
+    if (robot[key] in res) {
+      return null;
+    }
```
VinnieJ-2k20, line 33: For this task it is OK, but for the future it's better to use `hasOwnProperty` method.
`robot[key] also checks properties in parent objects.

VinnieJ-2k20 (approved):

PR #356, opened 2021-04-22
Solution

GoldenSpade:
[JS-file](https://github.com/BilliZane/js_inverse_robot/blob/develop/src/inverseRobot.js)

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -24,7 +24,28 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const mainPropArray = [];
+  const robotReverse = {};
+
+  for (const key in robot) {
+    mainPropArray.push(key, robot[key]);
+  }
+
+  mainPropArray.sort();
+
+  for (let i = 0; i < mainPropArray.length; i++) {
+    if (mainPropArray[i] === mainPropArray[i + 1]) {
+      return null;
+    }
+  }
+
+  Object.keys(robot).forEach(function(value) {
+    const key = robot[value];
+
+    robotReverse[key] = value;
+  });
+
+  return robotReverse;
 }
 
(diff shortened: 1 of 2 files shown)


stas-dukach (approved):
good

src/inverseRobot.js
```diff
@@ -24,7 +24,28 @@
...
+    mainPropArray.push(key, robot[key]);
+  }
+
+  mainPropArray.sort();
+
+  for (let i = 0; i < mainPropArray.length; i++) {
+    if (mainPropArray[i] === mainPropArray[i + 1]) {
+      return null;
+    }
+  }
```
VinnieJ-2k20, line 40: There are much more easier ways to check if an object has certain property.
Use `hasOwnProperty` here.

src/inverseRobot.js
```diff
@@ -24,7 +24,28 @@
...
+
+  Object.keys(robot).forEach(function(value) {
+    const key = robot[value];
```
VinnieJ-2k20, line 43: It's better to get used to arrow function from the start.

```suggestion
  Object.keys(robot).forEach((value) => {
    const key = robot[value];
```

VinnieJ-2k20 (changes requested):
You need to optimize your code.
Details in comments.

src/inverseRobot.js
```diff
@@ -23,8 +23,18 @@
...
  * @return {object}
  */
-function inverseRobot(robot) {
-  // write code here
-}
+
+const inverseRobot = (robot) => {
+  const robotReverse = {};
+
+  for (const key of Object.keys(robot)) {
```
```suggestion
  for (const key in robot) {
```
You could use `for`-`in` loop here

varseniuk (approved):
Good! 👍 
Left one comment

mykola-yarchuk (approved):

PR #360, opened 2021-04-23
Solution

MinchukO:


src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const rightRobot = {};
```
illisogori, line 28:
```suggestion
  const inversedRobot = {};
```

illisogori (approved):
Good Job.

src/inverseRobot.js
aliceblunt8: yes, inversedRobot sound better as a name for variable

aliceblunt8 (approved):
good job, keep going!

PR #362, opened 2021-04-23
ready for check

Bogdan-Kotsupey:


src/inverseRobot.js
```diff
@@ -23,8 +23,26 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const mirrorRobot = {};
+
+  for (const key in robot) {
+    if (mirrorRobot.hasOwnProperty(robot[key]) === true) {
```
aliceblunt8, line 31: hasOwnProperty returns true or false, no need to explicitly compare it to boolean

```suggestion
    if (mirrorRobot.hasOwnProperty(robot[key])) {
```

aliceblunt8 (approved):
good job!

PR #451, opened 2021-06-25
Solution

varopn:


src/inverseRobot.js
```diff
@@ -24,7 +24,11 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  return Object.entries(robot).reduce(
+    (accum, [key, value]) =>
+      !accum || accum[value] ? null : Object.assign(accum, { [value]: key })
+    , {}
+  );
```
vkryvytskyy, line 31: would be nice to solve this task without transforming object into array and back

vkryvytskyy (approved):

PR #454, opened 2021-06-26
Solution

ostapdev:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (robot[key] in inversedRobot) {
+      return null;
+    } else {
+      inversedRobot[robot[key]] = key;
+    }
```
vkryvytskyy, line 34:
```suggestion
    }
    
    inversedRobot[robot[key]] = key;
```
you don't need else, since there is return statement

vkryvytskyy (approved):

PR #506, opened 2021-08-02
Solution

alex-glotov67:


SashaLifashkin:
Not bad!

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (Object.keys(result).includes(robot[key])) {
```
```suggestion
    if (result[robot[key]]) {
```
This is a better-optimized solution

In your variant, you always iterate through the array, but in my variant, you just check if the object has certain value

varseniuk (changes requested):
Let's make it more optimized

varseniuk (approved):
👍

PR #511, opened 2021-08-02
Solution

denyschemeriakov:


ivanchenko-kirill (approved):
It`s really cool that you used Set in this task 👍

varseniuk (approved):
Clever.
But you could make it simply without `set` and just swap keys and values using `in` loop

PR #522, opened 2021-08-03
task is done

Oleksandr-Zhitenko:


src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newInverseRobot = {};
+
+  for (const key in robot) {
+    if (Object.keys(newInverseRobot).includes(robot[key])) {
```
alex-glotov67, line 31: you can use:
if (newInverseRobot[robot[key]]). Its work faster

alex-glotov67 (approved):
I like it. And i left my advice)

src/inverseRobot.js
Oleksandr-Zhitenko: ty

Clonvhs (approved):
very good 

that's my boy

danheim (approved):

PR #541, opened 2021-08-03
added solution

roman-hado:


Serhii-Usenko (approved):
nice)

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotReverse = {};
+
+  for (const key in robot) {
+    if (Object.keys(robotReverse).includes(robot[key])) {
```
```suggestion
    if (robotReverse[robot[key]]) {
```
This approach would be more optimized. Now you are iterating through the whole array to check if the key is present. In my example you just check one time if there is a key in the object

varseniuk (changes requested):
Good job, but let's make it more optimized

varseniuk (approved):

PR #542, opened 2021-08-03
first commit

VVHimiak:


maximbalyk (approved):
good job!

varseniuk (changes requested):
Let's try to make only one loop instead of two.
Use one `for (.... in ...)` loop and do everything in one iteration:
1. check if the key is currently present (then return null)
2. If the key is not present, add it with new value

varseniuk (approved):

PR #550, opened 2021-08-04
made inverse robot task

Uran-web:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (!fixedRobot.hasOwnProperty(robot[key])) {
+      fixedRobot[robot[key]] = key;
+    } else {
+      return null;
+    }
```
Ihor-Karpyn, line 34: Greate practice is write `if`  wich break your function and then main code
```suggestion
    if (fixedRobot[robot[key]]){
      return;
    }
    
    fixedRobot[robot[key]] = key;
```

Ihor-Karpyn (changes requested):
Almost done 👌

Ihor-Karpyn (changes requested):
Pay attention to prev comment

Ihor-Karpyn (approved):
LGTM

PR #584, opened 2021-09-14
solution

Bogdansgit:


Rocketjet (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const prop of Object.entries(robot)) {
+    if (prop[1] in newRobot) {
+      return null;
+    } else {
+      newRobot[prop[1]] = prop[0];
+    }
```
Anna-Tkachenko, line 34: else is redundant here
```suggestion
    newRobot[prop[1]] = prop[0];
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const prop of Object.entries(robot)) {
```
Anna-Tkachenko, line 29: Try to rewrite it without using Object.entries

Anna-Tkachenko (changes requested):

PR #586, opened 2021-09-15
Solution

zhenia-verbitskaya:


yungvitik:
Супер!

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
Anna-Tkachenko, line 27: inversedRobot name will be more meaningful
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      newRobot[robot[key]] = key;
+    }
```
Anna-Tkachenko, line 34: else is redundant here
```suggestion
    newRobot[robot[key]] = key;
```

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Well Done⭐️

PR #596, opened 2021-09-15
solution

AVronskyi:


Bogdansgit:
Nice job!

Bogdansgit (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
```
Anna-Tkachenko, line 27: Let's create more meaningful name for this variable instead of obj

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Well Done!⭐️

PR #599, opened 2021-09-15
solution

s-person27:


Denis-Demyanok (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const normalRobot = {};
```
Anna-Tkachenko, line 27: would be better named variable repairedRobot or inversedRobot

Anna-Tkachenko (approved):
Good job!👍

PR #602, opened 2021-09-15
solution

vlad-shulzhenko:


Denis-Demyanok (approved):

src/inverseRobot.js
```diff
@@ -23,8 +23,19 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+
+  for (const i of Object.entries(robot)) {
```
Anna-Tkachenko, line 31: Object.entries is redundant here
```suggestion
  for (const key of robot) {
```

Anna-Tkachenko (approved):

PR #613, opened 2021-09-16
Solution

YulyaVaV:


Mariia-Torkaienko (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverseObj = {};
```
Anna-Tkachenko, line 27: Don't use names of data types in variable name
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (inverseObj[robot[key]] in robot) {
+      return null;
+    }
```
Anna-Tkachenko, line 32: It's better to check if robot[key] value already exists in inversedRobot, not in robot
```suggestion
    if (inverseObj[robot[key]]) {
      return null;
    }
```

Anna-Tkachenko (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
```
Anna-Tkachenko, line 27: Don't use names of data types (obj, object, arr, array) in variable name❗️

Anna-Tkachenko (approved):
Good job👍

PR #614, opened 2021-09-16
Solution

DY-JS:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+    if (res.hasOwnProperty(key)) {
+      break;
+    }
+    res[robot[key]] = key;
+  }
 
+  if
+  (Object.values(res).length < robVal.length) {
+    return null;
+  }
```
aliceblunt8, line 40: you can return null from inside the loop, no need to compare values length
(use properly hasOwnProperty method)

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
-}
+  const robVal = Object.values(robot);
+  const res = {};
+
+  for (const key in robot) {
+    if (res.hasOwnProperty(key)) {
+      break;
+    }
+    res[robot[key]] = key;
+  }
```
aliceblunt8, line 35:
```suggestion
  for (const key in robot) {
    if (res.hasOwnProperty(robot[key])) {
      return null;
    }
    res[robot[key]] = key;
  }
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #626, opened 2021-10-04
Solution

cyb1ove:


TRSKZK (approved):
First of all I don't understand why did you use try catch block in this task, and overall solution is hard to read, but I wonder what will mentor say))

src/inverseRobot.js
```diff
@@ -25,6 +25,21 @@
  */
 function inverseRobot(robot) {
   // write code here
+  try {
```
VinnieJ-2k20, line 28: try/catch block here is useless, you shouldn't get any unexpected errors from this task.

src/inverseRobot.js
```diff
@@ -25,6 +25,21 @@
...
+      Object.entries(robot).reduce((newRobot, [value, key]) => {
+        if (newRobot.filter(x => x[0] === key).length) {
+          return null;
+        } else {
+          return newRobot.concat([
+            [key, value],
+          ]);
+        }
+      }, [])
+    );
```
VinnieJ-2k20, line 39: This solution may work, but it's too complicated.
Rewrite your solution using `for ... in` loop and create object using simple `{}`.

VinnieJ-2k20 (changes requested):
Your code needs some optimisation.

src/inverseRobot.js
```diff
@@ -25,6 +25,32 @@
...
+  //       } else {
+  //         return newRobot.concat([
+  //           [key, value],
+  //         ]);
+  //       }
+  //     }, [])
+  //   );
+  // } catch (e) {
+  //   return null;
+  // }
```
VinnieJ-2k20: Always delete commented code before committing.

src/inverseRobot.js
```diff
@@ -25,6 +25,32 @@
...
+  const resultRobot = {};
+
+  for (const key in robot) {
+    if (!Object.keys(resultRobot).includes(robot[key])) {
```
VinnieJ-2k20, line 30: Don't extract keys inside loop.
Do it once before and use variable.

VinnieJ-2k20 (changes requested):
This solution is much better.
Just some final changes left.

VinnieJ-2k20 (approved):
Good job 👍 
But your solution could be simplified even more if you just check
`if (value in resultRobot) {`

PR #628, opened 2021-10-04
solved

arielg0rl:


Olexandr6 (changes requested):
Your solution needs some optimisation.
You can solve this task just using `for (... in ...) {}` loop.

VinnieJ-2k20 (changes requested):
Agreed. Let's not use unnecessary arrays and nested loops.
Use `for ... in` loop to cycle through `robot` properties.

VinnieJ-2k20 (approved):
Good job 👍

PR #630, opened 2021-10-05
solution

dima-narozhnyi:


Olexandr6 (approved):
Well👍

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (!newRobot[robot[key]]) {
```
VinnieJ-2k20, line 30: If reversed robot will have this property, but it will be falsy - `''`, `0`, `null` or `undefined`, your code will work incorrectly.
Need to use another way to check if object has certain property.

VinnieJ-2k20 (changes requested):
Almost done.

VinnieJ-2k20 (approved):
Good job 👍

PR #631, opened 2021-10-05
solution

Olexandr6:


MariiaObraztsova (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const [value, key] of Object.entries(robot)) {
+    if (inversedRobot[key] === undefined) {
```
VinnieJ-2k20, line 30: What if it will store `undefined` ???
Use another check for property presence.

VinnieJ-2k20 (changes requested):
Solution looks good, but it needs one fix.

VinnieJ-2k20 (approved):
Good job 👍

PR #633, opened 2021-10-05
Solution.

AlinaKostiuk21:


denys-anosov (approved):
Nice solution! 👍

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    const resultKey = Object.keys(result);
+
+    if (resultKey.includes(robot[key])) {
+      return null;
+    }
```
VinnieJ-2k20, line 34: This can be simplified.
```suggestion
    if (robot[key] in result) {
      return null;
    }
```

VinnieJ-2k20 (approved):
Good job 👍 
Read the comment.

PR #641, opened 2021-10-06
Solution

Anton-Karpena:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    if (robot[key] in reversedRobot) {
+      return null;
+    }
+
+    reversedRobot[robot[key]] = key;
```
VinnieJ-2k20, line 34: You should've kept variable for robot[key], because it's used twice.

PR #643, opened 2021-10-06
solution

Linetstg:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const rever = {};
```
VinnieJ-2k20, line 27: Use more appropriate variable name. What's `rever` ?

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (!rever.hasOwnProperty(robot[key])) {
+      rever[robot[key]] = key;
```
VinnieJ-2k20, line 31: robot[key] is used twice. It's better to create a variable.

PR #679, opened 2021-11-08
Solution

ImIkator:
Inversed

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
+      return null;
+    }
+    result[robot[key]] = key;
+  }
+
+  return result;
 }
 
 module.exports = inverseRobot;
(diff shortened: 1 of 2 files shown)


src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
+      return null;
+    }
```
dmtrhrytsak, line 33: Add a space before if statement)
```suggestion
    }
    
```

src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const result = {};
+
+  for (const key in robot) {
```
dmtrhrytsak, line 30: You could also use Object.entries(robot) with destructuring here

```
   // your code
  }
```

dmtrhrytsak (approved):

src/inverseRobot.js
illiarerun: Yeah, as one of the ways to iterate through object

src/inverseRobot.js
illiarerun: I guess that Dima has meant that you should add a newline after!`if` clause to improve code readability

illiarerun (approved):
Good job 🔥  Check comments also:

PR #683, opened 2021-11-08
Task completed

Vitaliy-0:


JamesSunderlanddd (approved):
Good Job!

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
AlinaMatuschak, line 27: `result` isn't informative 
```suggestion
  const newRobot = {};
```
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (result[robot[key]]) {
```
AlinaMatuschak, line 30: Value can be 0. So you rewrite this value but should return null. Use hasOwnProperty )

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+      return null;
+    }
+
+    result[robot[key]] = key;
```

AlinaMatuschak (changes requested):
Good start 💪

AlinaMatuschak (approved):

PR #686, opened 2021-11-08
solution

Val-space:


src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+    for (let j = 0; j < values.length; j++) {
+      if (j !== i && values[i] === values[j]) {
+        return null;
+      }
+    }
+  }
+
+  for (const i in robot) {
+    inversedRobot[robot[i]] = i;
+  }
```
dmtrhrytsak, line 40: How about using for of loop with Object.entries combined in it?

```suggestion
for (const [key, value] of Object.entries(robot)) {
  // Check if value already exist as a property, if true, return null
  // Inverse => inversedRobot[value] = key
}

**Review by dmtrhrytsak — APPROVED** [review]

File: src/inverseRobot.js [hunk 2]
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (inversedRobot[value]) {
```
illiarerun, line 30: It'd be better to use `value in` or `hasOwnProperty()` to check if the key exists in the object. You got lucky that we don't have values like: `false`, `0` or empty string in this task, but your check wouldn't work properly with such values.

illiarerun (changes requested):
Great start 🔥

src/inverseRobot.js
dmtrhrytsak: `hasOwnProperty()` wouldn't pass the linter, so I guess using `in` operator is the right answer

Val-space: I guess linter ignores me..because hasOwnProperty() passed

illiarerun (approved):
LGTM 🔥

illiarerun (approved):

PR #689, opened 2021-11-08
Solution

msaikun:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const i in robot) {
+    if (result[robot[i]]) {
```
sviatoslav-myhal, line 30: can use hasOwnProperty, but honestly, i don`t even know, if it makes some sense)

sviatoslav-myhal (commented):
good job!)

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
volodymyr-soltys97, line 27: ```result``` isn't informative
```suggestion
  const newRobot = {};
```

volodymyr-soltys97, line 27:
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
volodymyr-soltys97, line 30: Value can be 0. So you rewrite this value but should return null. Use hasOwnProperty.

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
+
+  for (const i in robot) {
+    if (result[robot[i]]) {
+      return null;
+    }
+    result[robot[i]] = i;
```
volodymyr-soltys97, line 33: you can create constant for robot[і]

volodymyr-soltys97 (approved):
Good job 👍

AlinaMatuschak (approved):
🚀

PR #690, opened 2021-11-08
Solution

olya-shyian:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
```
dmtrhrytsak, line 27: Think of a better way to name your object, be more descriptive 👀

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (newObj[robot[key]]) {
```
dmtrhrytsak, line 30: What if the value turn out to be on of the falsy values, like 0 or null?)

```suggestion
    if (key in newObj) { // or use hasOwnProperty
```

dmtrhrytsak (approved):

src/inverseRobot.js
AlinaMatuschak: Agree. Value can be 0. So you rewrite this value but should return null. Use hasOwnProperty )

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+      return null;
+    }
+
+    newObj[robot[key]] = key;
```

AlinaMatuschak (changes requested):
Almost perfect 👍

AlinaMatuschak (approved):
🚀

PR #692, opened 2021-11-08
Solution

volodymyr-soltys97:


tmrsl (approved):
Looks good.

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
AlinaMatuschak, line 27: `result` isn't informative 
```suggestion
  const newRobot = {};
```
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (result[robot[key]]) {
```
AlinaMatuschak, line 30: Value can be 0. So you rewrite this value but should return null. Use hasOwnProperty )

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+      return null;
+    }
+
+    result[robot[key]] = key;
```

AlinaMatuschak (changes requested):
Good start 👌

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const robotKey = robot[key];
+
+    if (robotKey !== 0 && inversedRobot.hasOwnProperty(robotKey)) {
```
AlinaMatuschak, line 32:
```suggestion
    if (inversedRobot.hasOwnProperty(robotKey)) {
```

AlinaMatuschak (approved):

PR #697, opened 2021-11-08
Yep thats the right one (I suggest)

skachkov23:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobots = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (value in inversedRobots) {
```
volodymyr-soltys97, line 30: you can use hasOwnProperty

volodymyr-soltys97 (approved):
GJ

src/inverseRobot.js
AlinaMatuschak: Agree. Use `hasOwnProperty` instead `in`

AlinaMatuschak (approved):
nice 👍

src/inverseRobot.js
skachkov23: Thank you, resolve)

PR #698, opened 2021-11-08
solution

mariia-bila:


src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
...
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    const invKey = [robot[key]];
+
+    inverse[invKey] = key;
+  }
```
dmtrhrytsak, line 42: You could rewrite it using **for of loop** in conjunction with **Object.entries**

```suggestion
for (const [key, value] of Object.entries(robot)) {
  // Check if value already exist as a property, if true, return null
  // Inverse => inverse[value] = key
}
```

dmtrhrytsak (approved):

src/inverseRobot.js
illiarerun: Reasonable comment, AGREE

src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
```
illiarerun, line 27: Be more specific with variable names, something like `inversedRobot` would be a perfect fit here

illiarerun (changes requested):
Great start 🔥  Let's work on fixes

illiarerun (changes requested):

illiarerun (approved):
LGTM 🔥

PR #699, opened 2021-11-09
solution

DudarDaryna:


dlugash (approved):
Wow! Looks very good for me!

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (newRobot.hasOwnProperty(value)) {
+      return null;
+    } else {
+      newRobot[value] = key;
+    }
```
illiarerun, line 34: You should avoid unnecessary nesting, you can return null in inverted if block and assign values after that block

illiarerun (changes requested):
Great start 🔥

illiarerun (approved):
LGTM 🔥

PR #707, opened 2021-11-09
Solution

dlugash:


junglq (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const obj = robot;
```
illiarerun, line 27: Redundant, threre is no need to make this reference to an object
```suggestion
```

src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
...
+    newRobot[newKey] = oldKey;
+  }
+
+  const size1 = Object.keys(robot).length;
+
+  const size2 = Object.keys(newRobot).length;
+
+  if (size1 !== size2) {
+    return null;
+  }
```
illiarerun, line 42: All these checks are after the copy, you can improve your code by checking if the key is repeated right in the loop above, a little hint: `key in` or `hasOwnProperty()` would do the trick)

illiarerun (changes requested):
Good start 🔥

illiarerun (approved):
LGTM 🔥

PR #711, opened 2021-11-10
Solution

IrynaHerasymchuk:


src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+  const uniqueRobotValues = [...new Set(robotValues)];
+
+  if (robotValues.length !== uniqueRobotValues.length) {
+    return null;
+  }
```
Nadiyahr, line 32: You could have checked for duplicates with  ` if (fixedRobot.hasOwnProperty(robot[key])) {
      return null;
    }` inside the loop for, it would be shorter. But you solution also works.

Nadiyahr (approved):

src/inverseRobot.js
IrynaHerasymchuk: > You could have checked for duplicates with ` if (fixedRobot.hasOwnProperty(robot[key])) { return null; }` inside the loop for, it would be shorter. But you solution also works.

You're right, this solution would be simpler! Thanks for pointing this out😄

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+  const uniqueRobotValues = [...new Set(robotValues)];
```
illiarerun, line 28: There's no need to store Set as an array, if you use it only for comparing length, you can use `size` property

illiarerun (changes requested):
Good job 🔥  But let's fix one thing:

illiarerun (approved):
LGTM 🔥

PR #736, opened 2021-12-13
done

Niki-veb-dev:


Rmu4h (approved):

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const values = Object.entries(robot);
+  const newRobot = {};
+
+  for (const v of values) {
```
Denchepornyuk, line 31: `v` isn't a good name for variable

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
...
+    if ([v[1]] in newRobot) {
```
Denchepornyuk, line 32: `hasOwnProperty` is better for checking keys in object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #739, opened 2021-12-13
solution

Toros-Zz:


src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const keys = [];
+  const prop = [];
```
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
...
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (keys.includes(robot[key])) {
```
Olexandr6, line 32: use `in` operator, instead

src/inverseRobot.js
Olexandr6: Don't use additional arrays, you can do the same just with a new object

Olexandr6 (changes requested):
Almost done

Olexandr6 (approved):
Good solution 👍

PR #749, opened 2021-12-14
Solution

Rom911:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+
+  const newRobot = {};
+
+  for (let i = 0; i < values.length; i++) {
+    if (values.indexOf(values[i]) !== values.lastIndexOf(values[i])) {
```
mykola-koval, line 32: можно использовать цикл без индексов

mykola-koval (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+
+  const newRobot = {};
+
+  for (let i = 0; i < values.length; i++) {
```
Denchepornyuk, line 31: It's better to use `for in` loop for object, and Object.values is unnecessary here

src/inverseRobot.js
Denchepornyuk, line 32: use `hasOwnProperty` for checking if keys exist in the object

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+    if (values.indexOf(values[i]) !== values.lastIndexOf(values[i])) {
+      return null;
+    }
+  }
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
```
Denchepornyuk, line 39: you can do this task only with one loop `for`

Denchepornyuk (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
...
+  const rightRobot = {};
+
+  for (const detail in robot) {
+    if (rightRobot.hasOwnProperty(robot[detail])) {
+      return null;
+    }
+    rightRobot[robot[detail]] = detail;
+  }
+
+  return rightRobot;
```
Denchepornyuk, line 45: this loop is the same as the previous one, you can remove it and return `inversed` from the function

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #750, opened 2021-12-14
Solution

ErnestoFernandezUA:


aleksandra-mileant (approved):

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const temp = {};
+
+  for (const b in robot) {
```
Denchepornyuk, line 30: `b` isn't a good name for the variable

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
...
+    if (Object.keys(temp).includes(robot[b])) {
```
Denchepornyuk, line 31: you can use `hasOwnProperty` here

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
...
+      return null;
+    }
+
+    temp[robot[b]] = b;
+    delete robot[b];
+  }
+
+  for (const key in temp) {
+    robot[key] = temp[key];
+  }
```
Denchepornyuk, line 41: You don't need to change robot object

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
...
+
+  return robot;
```
Denchepornyuk, line 43: you should return new object, not `robot`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #751, opened 2021-12-14
Develop

bcmp4p1:


destroythemoon (approved):
LGTM!

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotInverse = {};
+  const entries = Object.entries(robot);
+
+  for (let n = 0; n < entries.length; n++) {
+    for (let i = n + 1; i < entries.length; i++) {
```
Denchepornyuk, line 31: don't write loop inside another loop
This task you can do only with one loop for

Hint: for checking if key exists in the object use `hasOwnProperty`

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotInverse = {};
+  const entries = Object.entries(robot);
```
Denchepornyuk, line 28: Object.entries is redundant, you can use loop `for in` for the object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #753, opened 2021-12-14
Solution

Bohdan-Maksymiuk:


HrynykNataliia (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]]) {
```
Denchepornyuk, line 30: better to use `hasOwnProperty` for checking if keys exist in the object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #771, opened 2021-12-15
solution

andrii-koptiev:


luckdenny (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,31 @@
...
+  }
+
+  return inversed;
+}
+
+function findDuplicates(array) {
+  const noDuplicatesArray = Array.from(new Set(array));
+
+  return noDuplicatesArray.length === array.length;
 }
```
Denchepornyuk, line 52: Very complicated logic.
You need to create `for in` loop for robot object and inside loop you can check if key exists in object `inversed` or not. If exist then return `null` if not add this key to `inversed` object

Denchepornyuk (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const values in robot) {
+    if (inversed[values] || inversed[robot[values]]) {
```
Denchepornyuk, line 30: better to use hasOwnProperty here

Denchepornyuk (approved):

PR #794, opened 2022-01-09
solution [Yuriy Shvorob]

djtyrf312:


src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+    for (let j = i + 1; j < valArray.length; j++) {
+      if (valArray[j] === valArray[i]) {
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    resultObj[robot[key]] = key;
+  }
```
aliceblunt8, line 41: I believe you only need one loop in this task (for in)

aliceblunt8: also use hasOwnProperty to check whether there's certain property in object

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const resultObj = {};
+
+  const valArray = Object.values(robot);
```
aliceblunt8, line 29: remove, iterate through object with for in loop
```suggestion
```

aliceblunt8 (changes requested):

PR #796, opened 2022-01-11
Solution

k-romakovalchuk:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    inversedRobot[robot[key]] = key;
+  }
+
+  if (Object.keys(inversedRobot).length !== Object.keys(robot).length) {
```
aliceblunt8, line 33: use hasOwnProperty here

aliceblunt8: place this condition inside the loop

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #802, opened 2022-01-18
)

tvoyammn:


krasnobokydima (approved):
Good

src/inverseRobot.js
```diff
@@ -24,7 +24,15 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const res = {};
```
Anna-Tkachenko, line 27: Let's create more meaningful variable name instead of res

src/inverseRobot.js
```diff
@@ -24,7 +24,15 @@
...
+
+  Object.keys(robot).forEach(key => {
+    res[robot[key]] = key;
+  });
```
Anna-Tkachenko, line 31: 👍

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #814, opened 2022-01-20
Solution

oleksandr-kozhushko:


krasnobokydima (approved):
Good!

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
```
```suggestion
  const reversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (reverseRobot[robot[key]]) {
```
Anna-Tkachenko, line 30: Check point 2 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #847, opened 2022-02-08
Inverted key-values

aspurgilus:


Ihor-Karpyn (changes requested):
Let's simplify this solution

Ihor-Karpyn (approved):
Good job

PR #848, opened 2022-02-08
solution for #inverseRobot

jstmpelowycz:


src/inverseRobot.js
```diff
@@ -13,18 +13,41 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+  const values = Object.values(robot);
+
+  if (values.some(function(V, index) {
+    return values.indexOf(V) !== index;
+  })) {
+    return null;
+  }
```
Ihor-Karpyn, line 44:
```suggestion
	const hasDuplicate = values.some((V, index) => (
   values.indexOf(V) !== index;
  ))

  if (hasDuplicate) {
    return null;
  }
```

Ihor-Karpyn (changes requested):
Almost done

src/inverseRobot.js
```diff
@@ -13,18 +13,43 @@
...
+
+  const hasDuplicate = values.some((V, index) => {
+    return values.indexOf(V) !== index;
+  });
+
+  if (hasDuplicate) {
+    return null;
+  }
+
+  for (const [K, V] of Object.entries(robot)) {
```
Ihor-Karpyn, line 48: better to use for in

src/inverseRobot.js
```diff
@@ -13,18 +13,43 @@
...
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+  const values = Object.values(robot);
+
+  const hasDuplicate = values.some((V, index) => {
+    return values.indexOf(V) !== index;
+  });
```
Ihor-Karpyn, line 42:
```suggestion
```

this code add 3 redundant iteration

src/inverseRobot.js
Ihor-Karpyn: can check inside loop 
if (key in inverse) return null

Ihor-Karpyn (changes requested):
Let's simlify solution

Ihor-Karpyn (approved):
Good job

PR #850, opened 2022-02-08
solution

SanyaBratashchuk:


Marta1993-droid (approved):

Ihor-Karpyn (changes requested):
---

Ihor-Karpyn (changes requested):
Simplify this solution, .values and entries redundant here. Use for in loop.

Ihor-Karpyn (approved):

PR #888, opened 2022-03-17
Solution

serhii56465:


Nazar-kob (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
-  // write code here
+  const sett = new Set();
+  const copyy = {};
+
+  for (const x in robot) {
+    sett.add(robot[x]);
+  }
+
+  if (sett.size !== Object.keys(robot).length) {
+    return null;
```
MasterpieceElbow, line 35: I think using Set() and this check is redundant. You can check if there is an existing key while adding a key.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #891, opened 2022-03-22
solution

TetianaKoval:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
 function inverseRobot(robot) {
-  // write code here
+  let newRobot = {};
+
+  for (const key in robot) {
+    if (!newRobot[robot[key]]) {
+      newRobot[robot[key]] = key;
+    } else {
+      newRobot = null;
+      break;
```
marina-tilniak, line 34:
```suggestion
      return null;
```

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+    };
+  }
+
+  return (newRobot);
```
marina-tilniak, line 38:
```suggestion
  return newRobot;
```

src/inverseRobot.js
marina-tilniak, line 34: swap if branches, because its better to have condition without not if you have else

marina-tilniak (dismissed):

PR #894, opened 2022-03-23
'Solution'

denyslipin:


serhii56465 (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    newRobot[value] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(newRobot).length) {
+    return null;
+  }
```
MasterpieceElbow, line 37: That is not an obvious check for duplicated keys, better check it while adding in the loop.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #899, opened 2022-03-28
Solution

yaremenko-maksym:


Anna-Tkachenko (changes requested):
Good job!
Just try to optimize your code, use 1 iteration instead of 3: Object.values (1 iteration), Object.keys (2 iteration), for (3 iteration)

Anna-Tkachenko (approved):
Good job 👍

PR #900, opened 2022-03-29
solution

vladyslava-buzova:


src/inverseRobot.js
```diff
@@ -23,8 +23,24 @@
...
+      }
+    }
+  }
+
+  const swapped = Object.fromEntries(
+    // преобразовать в массив, затем map, затем fromEntries обратно объект
+    Object.entries(robot).map(([key, value]) => [value, key])
+  );
+
+  return swapped;
```
Bohdan-Maksymiuk, line 43: So many loops and iterations 😟 
You can just
* create new object (`newRobot` for example )
* iterate through keys (use `for ... in`)
* in loop check if newRobot has property `robot[key]` -> return null
* in another case add new property to newRobot

Bohdan-Maksymiuk (changes requested):

src/inverseRobot.js
vladyslava-buzova: Thank you :)

marina-tilniak (approved):

PR #905, opened 2022-03-29
Robot

MariaBondar:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+  const keys = Object.keys(robot);
+  const res = {};
```
YegorKochetkov, line 29: Checklist:

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+
+  for (let i = 0; i < values.length; i++) {
+    if (values[i] in res) {
+      return null;
+    }
```
YegorKochetkov, line 34: your solution passes the checklist, but it's better to use native object methods
Checklist:
GOOD EXAMPLE:

if (key in robot) {
EVEN BETTER EXAMPLE:

YegorKochetkov (changes requested):
Help to **repair** the **inverse robot** ;)

Anna-Tkachenko (approved):
Good job! 👍 
Try to optimize your code using only one iteration, now you use 3: Object.values (1), Object.keys (2), for (3)

PR #908, opened 2022-03-30
Inverse robot function was created

OksanaTyshchenko:


src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
+    const newValue = key;
+
+    if (inverse[newKey] !== undefined) {
+      return null;
+    } else {
```
yaremenko-maksym, line 35: no else needed. You already and up function, if inverse[newKey] !== undefined.

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
+    const newValue = key;
```
yaremenko-maksym, line 31: You can use Object.keys / Object.values instead of this loop. And this is better for me not to declare variables inside the loop.

yaremenko-maksym (changes requested):
Might use some minor fixes

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
```
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  for (const key in robot) {
+    const newKey = robot[key];
+    const newValue = key;
+
+    if (inverse[newKey] !== undefined) {
```
Anna-Tkachenko, line 33: Check point 2 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job 👍

PR #910, opened 2022-03-30
solution

HelenPetrovska:


src/inverseRobot.js
```diff
@@ -23,8 +23,21 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
   // write code here
+  const result = {};
```
YegorKochetkov, line 29: Checklist:

YegorKochetkov (changes requested):
Help to **repair** the **inverse** **robot** ;)

Anna-Tkachenko (approved):
Good job!
Just try to optimize your code using only 1 iteration instead of 3 (Object.keys, Object.values, for cycle)

PR #914, opened 2022-03-30
solution

Jenia80:


src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
 function inverseRobot(robot) {
-  // write code here
+  let result = {};
+  const check = [...new Set(Object.values(robot))];
+
+  if (check.length === 0) {
+    result = {};
+  } else if (check.length !== Object.values(robot).length) {
+    return null;
+  } else {
```
YegorKochetkov, line 34: [CODE STYLE] - You should return the value immediately in if statements. And don't use else after return statement. If return is performed - code after it won't be executed anyway.

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let result = {};
+  const check = [...new Set(Object.values(robot))];
+
+  if (check.length === 0) {
+    result = {};
+  } else if (check.length !== Object.values(robot).length) {
```
YegorKochetkov, line 32: your solution is too complicated. to check that the object already has such a key, it is better to use hasOwnProperty:
[CODE STYLE] - use correct check for property presence in object. Some properties can be present, but still contain falsy value.
GOOD EXAMPLE:
if (key in robot) {

EVEN BETTER EXAMPLE:

if (result.hasOwnProperty(key))

YegorKochetkov (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
Anastasiia-Sharkova, line 27: [NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)

Anastasiia-Sharkova (approved):

PR #916, opened 2022-03-30
solution

OleksandrNi:


Anastasiia-Sharkova (approved):

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const result = {};
```
Anna-Tkachenko, line 28: Will be better to create more meaningful variable name

Anna-Tkachenko (approved):
Good job 👍

PR #917, opened 2022-03-30
Solution of Invertion

maxxnikiforov:


src/inverseRobot.js
```diff
@@ -24,7 +24,29 @@
...
+      if (val[i] === val[x]) {
+        counter++;
+      }
+    }
+
+    if (counter > 1) {
+      return null;
+    }
+    counter = 0;
+  }
```
YegorKochetkov, line 43: nested loop is evil
checklist:
```
GOOD EXAMPLE:
if (key in robot) {

EVEN BETTER EXAMPLE:
```
use this constructions to return `null` where you iterate over the keys

YegorKochetkov (changes requested):
too confusing, could be much easier to do

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const invers = {};
```
Anna-Tkachenko, line 27: What is invers? Create more meaningful variable name

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+  const keys = Object.keys(robot);
+  const val = Object.values(robot);
+
+  for (let i = 0; i < keys.length; i++) {
```
Anna-Tkachenko, line 31: You can use for (const key of robot) cycle instead of creating two separate arrays with keys and values

Anna-Tkachenko (approved):

PR #920, opened 2022-03-31
Develop

Yurii837:


src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
...
+
+  for (let i = 0; i < arr.length; i++) {
+    const shifted = arr.shift();
+
+    if (arr.includes(shifted)) {
+      return null;
+    }
+  }
+
+  const result = {};
```
Anastasiia-Sharkova, line 39: [NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)

Anastasiia-Sharkova (changes requested):

Anna-Tkachenko (changes requested):
Your code works, but you can do this task using only 1 iteration
Now you use at least 6 iterations (Object.keys, Object.values, destructuring, includes, two for cycles)
Try to optimize your code and complete task in one iteration

Anna-Tkachenko (changes requested):
So many iterations..try to optimize your code
You can complete this task using only one cycle
You should only add needed properties to the new object and check, if this property is already existed in your object

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverseObject = {};
+
+  for (const key in robot) {
+    if (inverseObject[robot[key]]) {
+      return null;
+    }
+    inverseObject[robot[key]] = key;
```
MaksimusCRS, line 33:
```suggestion
    }
    
    inverseObject[robot[key]] = key;
```

MaksimusCRS (approved):

PR #926, opened 2022-03-31
Solution

AntonLevchuk:


src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+  const objKeys = Object.keys(robot);
+  const objValues = Object.values(robot);
+
+  const newObj = {};
+
+  const objSet = new Set(objValues);
+
+  if (objValues.length !== objSet.size) {
+    return null;
+  }
```
IlliaBezditnyi, line 36: ### You should check it with hasOwnProperty() according to checklist

IlliaBezditnyi, line 36: You should check it with hasOwnProperty() according to checklist
```suggestion
```

IlliaBezditnyi (changes requested):
Few changes

IlliaBezditnyi (approved):

Anna-Tkachenko (approved):
You can make your code easier and do this task in 1 iteration instead of 3

PR #930, opened 2022-03-31
add task

Helexi:


src/inverseRobot.js
```diff
@@ -23,8 +23,21 @@
...
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const valueObj = Object.values(robot);
+  const keyObj = Object.keys(robot);
+  const copy = {};
+
+  for (let i = 0; i < valueObj.length; i++) {
```
YegorKochetkov, line 32: much easier that way
```suggestion
  const copy = {};

  for (const key in robot) {
```

src/inverseRobot.js
```diff
@@ -23,8 +23,21 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const valueObj = Object.values(robot);
+  const keyObj = Object.keys(robot);
+  const copy = {};
```
YegorKochetkov, line 30: from checklist:
maybe, repairRobot as opposite to inverseRobot would be better

src/inverseRobot.js
```diff
@@ -23,8 +23,21 @@
...
+
+  for (let i = 0; i < valueObj.length; i++) {
+    copy[valueObj[i]] = keyObj[i];
+  }
+
+  if (Object.values(copy).length === Object.values(robot).length) {
+    return copy;
+  } else {
+    return null;
+  }
```
YegorKochetkov, line 40: from checklist:
GOOD EXAMPLE:

if (key in robot) {

EVEN BETTER EXAMPLE:

YegorKochetkov (changes requested):
if follow the checklist, then should redo

src/inverseRobot.js
```diff
@@ -23,8 +23,18 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const repairRobot = {};
```
Anna-Tkachenko, line 28:
```suggestion
  const repairedRobot = {};
```

Anna-Tkachenko (approved):
Good job! 👍

PR #931, opened 2022-03-31
solution

kkiirill:


SeKaM0 (approved):
Looks good

Anna-Tkachenko (approved):
Good job!
Just try to optimize your code using only 1 iteration instead of 3 (Object.keys, Object.values, for cycle)

PR #935, opened 2022-03-31
Solution

AmbeMO:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const copyObj = {};
```
Anna-Tkachenko, line 27: Will be better to create more meaningful variable name

Anna-Tkachenko (approved):
Good job! 👍

PR #937, opened 2022-03-31
ready

Nikitiliok:


viacheslav-petrovskyi (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,27 @@
...
+      if (robot[key] === robot[item]) {
+        counter++;
+      }
+    }
+
+    if (counter > 1) {
+      return null;
+    }
+    counter = 0;
+  }
```
Anna-Tkachenko, line 41: This cycle in redundant in your code
You can if statement for the next cycle and check if the property is already exists in robot - return null

Anna-Tkachenko (approved):

PR #938, opened 2022-03-31
Solution

serhii-naumenko:


Anastasiia-Sharkova (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
-  // write code here
+  const newRobot = {};
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+
+  if (keys.length === 0) {
+    return {};
+  }
+
+  for (let i = 0; i < keys.length; i++) {
```
Anna-Tkachenko, line 35: You can also use for (const key in robot) cycle instead of creating keys and values arrays

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+    if (newRobot.hasOwnProperty(values[i])) {
+      return null;
+    } else {
+      newRobot[values[i]] = keys[i];
+    }
```
Anna-Tkachenko, line 40: else is redundant here
```suggestion
    if (newRobot.hasOwnProperty(values[i])) {
      return null;
    } 
      
    newRobot[values[i]] = keys[i];
```

PR #944, opened 2022-04-02
Solution

alexandr1223:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
```
Bohdan-Maksymiuk, line 27: don't use obj, arr and etc. naming for variables.

Bohdan-Maksymiuk (approved):

PR #946, opened 2022-04-05
Solution

AndriiRut:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  for (const key in robot) {
+    if (robot[robot[key]]) {
```
Bohdan-Maksymiuk, line 28: better to use hasOwnProperty for checking if keys exist in the object

Bohdan-Maksymiuk (changes requested):
You should create a new object, instead of mutating one which is passed into arguments

Denchepornyuk (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    newRobot[robot[key]] = key;
+
+    delete newRobot[key];
```
Denchepornyuk, line 36: do you need `delete` here?

Denchepornyuk (changes requested):

src/inverseRobot.js
AndriiRut: I realized! I no longer need the`delete` because my new object was empty!

Denchepornyuk (approved):

PR #947, opened 2022-04-06
solved

misha200119:


danylo-onishchuk (changes requested):
use const key in robot
you need no Object.keys(robot) and Object.values(robot)

danylo-onishchuk (approved):

PR #948, opened 2022-04-07
solution inverseRobot task

koshachina:


src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
...
 function inverseRobot(robot) {
   // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (!invertedRobot.hasOwnProperty(robot[key])) {
+      invertedRobot[robot[key]] = key;
+    } else {
+      return null;
+    }
```
ArtemTopolskyi, line 35: you can return null if statement is truth and then just write main logic in the end of loop - as a way to avoid 'else' statement

ArtemTopolskyi (approved):

PR #956, opened 2022-04-10
Solution

alexmay112:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const keysPrev = [];
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (keysPrev.includes(robot[key])) {
```
marina-tilniak, line 31: you can use hasOwnProperty here, then  will not need array keyPrev

marina-tilniak (changes requested):

ArturErmolenko (approved):
well done 👍

PR #971, opened 2022-04-20
solution

efdalyalcin:


iLebo-stack (approved):
Good job Efdal👍

src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
...
-  // write code here
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotValues.length; i++) {
+    for (let j = i + 1; j < robotValues.length; j++) {
+      if (robotValues[i] === robotValues[j]) {
+        return null;
+      }
+    }
+  }
```
VinnieJ-2k20, line 35: 1. Let's avoid nested loops as much as possible.
2. Let's perform this check alongside adding properties to new object.

VinnieJ-2k20 (changes requested):
Your code works, but it can be optimized greatly.

src/inverseRobot.js
```diff
@@ -24,7 +24,24 @@
...
+
+    robotValues.push(robot[key]);
+  }
+
+  const robotKeys = Object.keys(robot);
+  const invRobot = {};
+
+  for (let i = 0; i < robotValues.length; i++) {
+    invRobot[robotValues[i]] = robotKeys[i];
+  }
```
VinnieJ-2k20, line 42: Avoid connecting two separate arrays logically here.
If any of them changes between initialization and usage - your program can be broken.
Having keys you can access corresponding values.

VinnieJ-2k20 (changes requested):
One change left.

meforoma (approved):

PR #973, opened 2022-04-20
Task completed

andreisucman:


src/inverseRobot.js
```diff
@@ -23,8 +23,20 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
+
+  if (Array.from(new Set(Object.values(robot))).length
+    < Object.values(robot).length) {
```
59MrRobot, line 31: If its possible, I'd suggest putting some parts of this in variables because it's kind of difficult to read as it is

59MrRobot (changes requested):
I've only got one comment, otherwise, good work!

src/inverseRobot.js
andreisucman: True 👍

59MrRobot (approved):
Good job

src/inverseRobot.js
```diff
@@ -23,8 +23,22 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
+  const lengthOfUniquePartsArray
+    = Array.from(new Set(Object.values(robot))).length;
```
VinnieJ-2k20, line 30: You can create `new Set` and then take it's `size` instead of Array `length`

VinnieJ-2k20 (approved):
Good job 👍 
Added one comment.

PR #975, opened 2022-04-21
Solution

amitpatiljc:


59MrRobot (approved):
Good job!

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+  const uniqueValues = Array.from(new Set(values)).length;
```
VinnieJ-2k20, line 30: You can read Set size instead of array length.

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
```
VinnieJ-2k20, line 29: Read only using keys, don't rely on elements order in two independent arrays (`keys` and `values`).

```suggestion
  const keys = Object.keys(robot);
```

VinnieJ-2k20 (changes requested):
Solution works, but it can still be optimized.

VinnieJ-2k20 (approved):
Good job 👍

PR #976, opened 2022-04-21
solution

iLebo-stack:


src/inverseRobot.js
```diff
@@ -24,7 +24,34 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const arrayOfProperties = [];
+  const arrayOfValues = [];
+  const obj = {};
+
+  for (const prop in robot) {
+    arrayOfValues.push(prop);
+    arrayOfProperties.push(robot[prop]);
+  }
```
59MrRobot, line 34: you can use Object.keys(object) and Object.values(object) to get an array of the keys and values of an object instead of manually doing it

src/inverseRobot.js
```diff
@@ -24,7 +24,34 @@
...
+
+  for (const propertyTracker of arrayOfProperties) {
+    let count = 0;
+
+    for (const property of arrayOfProperties) {
+      if (propertyTracker === property) {
+        count++;
+      }
+    }
```
59MrRobot, line 43: I would avoid using nested loops as much as possible so I suggest using the hasOwnProperty() method that we learnt in today's lesson to look for repeated properties.

59MrRobot (changes requested):
Good work, I've got a couple of suggestions to optimize your work

59MrRobot (approved):
Good work

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const arrayOfProperties = Object.values(robot);
+  const obj = {};
```
VinnieJ-2k20, line 28: I would use more definitive name, like `inversedRobot`.

VinnieJ-2k20 (approved):
Good job 👍 
Added one comment.

PR #978, opened 2022-04-22
Solution

huberciak:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
```
rHashTag, line 27: Perhaps use a variable name that's more obvious. The function outputs an inverse rather than a copy.

rHashTag (commented):
Good job on the solution!

src/inverseRobot.js
huberciak: +1

VinnieJ-2k20 (approved):
Good job 👍

PR #982, opened 2022-04-25
so messy solution

ZUBOV-ILLIA:


EnjSadman (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const len = Object.keys(robot).length;
+  const myObj = { ...robot };
```
Denchepornyuk, line 28: create an empty object, not a copy of `robot`

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+
+  for (const key in robot) {
+    delete robot[key];
+  }
```
Denchepornyuk, line 32: remove this loop, with `myObj = {}` you don't need this loop

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+
+  for (const key in myObj) {
+    robot[myObj[key]] = key;
+  }
+
+  return Object.keys(robot).length === len ? robot : null;
```
Denchepornyuk, line 38: check if the key exists in `myObj` using `hasOwnProperty` and if exist return `null` if not add this key to `myObj`

and return from the function `myObj` instead of `Object.keys(robot).length === len ? robot : null;`

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #983, opened 2022-04-26
inverse_robot

elkhayate:


src/inverseRobot.js
```diff
@@ -25,6 +25,23 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const output = robot;
```
meforoma, line 28: you don't create a copy, but just make a reference to same object `robot` - does not provide any value.

src/inverseRobot.js
```diff
@@ -25,6 +25,23 @@
...
+  const values = Object.values(output);
+  const result = {};
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = 0; j < values.length; j++) {
+      if (values[i] === values[j] && i !== j) {
+        return null;
+      }
+    }
+  }
```
meforoma, line 38: you could replace this whole code with a neat `result.hasOwnProperty(robot[key])` validation and returning null inside `for` loop

src/inverseRobot.js
```diff
@@ -25,6 +25,23 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const output = robot;
+  const values = Object.values(output);
```
meforoma, line 29: don't need this as well if properly use `hasOwnProperty` for validation

meforoma (changes requested):

src/inverseRobot.js
elkhayate: but i need to check if the value exist not the key

src/inverseRobot.js
```diff
@@ -25,6 +25,22 @@
...
+    for (let j = 0; j < values.length; j++) {
+      if (values[i] === values[j] && i !== j) {
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    result[robot[key]] = key;
+  }
```
VinnieJ-2k20, line 41: Delete nested loop above and put the check for duplicates here in this loop.

VinnieJ-2k20 (commented):
Almost done, final improvement left.

VinnieJ-2k20 (changes requested):
misclick

ihor-jpeg (approved):
well done!

ihor-jpeg (approved):

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

meforoma (approved):

PR #987, opened 2022-04-26
inverse key with value

Andriyk0:


SavMary (approved):
exellent)

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const valueArr = Object.keys(robot);
+  const keyArr = Object.values(robot);
+  const dublicate = keyArr.some((e, i, arr) => arr.indexOf(e) !== i);
```
Denchepornyuk, line 29: remove all these variables and use `hasOwnProperty` inside loop for instead of `some` method

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #992, opened 2022-04-26
Solution

PRO-GRAM-MER:


meforoma (changes requested):
hmm, solution works, but it's over-complicated.
you have an object and you can iterate through it with a single for cycle:
```
const inversed = {};

  // feed `robot[property] = property` to inversed
  // using Object.assign or square brackets `inversed[robot[property]] = property`
}
```

and `inversed.hasOwnProperty(robot[property])` inside `for` loop will help you validate to return null

src/inverseRobot.js
```diff
@@ -25,6 +25,37 @@
...
+  //   }
+  // }
+
+  // const swap = Object.entries(robot).map(
+  //   ([key, value]) => ({ [value]: key })
+  // );
+
+  // const newObject = Object.assign({}, ...swap);
+
+  // return newObject;
```
VinnieJ-2k20, line 44: Don't keep commented code. If you don't need it  - delete it.

VinnieJ-2k20 (approved):
Good job 👍

PR #993, opened 2022-04-26
inverseRobot

Oleksandr-Ivanchenko:


valerii-derkach (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversion = {};
+
+  for (const key in robot) {
+    inversion[robot[key]] = key;
+  }
+
+  if (Object.keys(inversion).length !== Object.keys(robot).length) {
```
Denchepornyuk, line 33: use `hasOwnProperty` to check if a key exists in object

Denchepornyuk (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversion = {};
```
```suggestion
  const inversedRobot = {};
```

Anna-Tkachenko (approved):

PR #998, opened 2022-04-26
First_commit_InverseRobot

igoree88:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const arr = [];
```
iwanttosmile4u, line 28: This array is redundant.

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+
+  for (const key in robot) {
+    if (arr.includes(robot[key])) {
```
iwanttosmile4u, line 31: You can use hasOwnProperty in this case to check if robot value repeat

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+      return null;
+    }
+    arr.push(robot[key]);
```
iwanttosmile4u, line 34: This is not a necessary operation also. You can delete it.

src/inverseRobot.js
igoree88: Fixed

src/inverseRobot.js
igoree88: Fixed. I forgot about this method when I was writing code))

src/inverseRobot.js
igoree88: Fixed

Denchepornyuk (approved):

PR #999, opened 2022-04-26
Develop

MabuzaM:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const part in robot) {
+    if (newRobot.hasOwnProperty(robot[part])) {
+      return null;
+    } else {
+      newRobot[robot[part]] = part;
```
59MrRobot, line 33: There's no need to wrap this line in an else block

59MrRobot (changes requested):

src/inverseRobot.js
MabuzaM: Okay noted. Have fixed please check again

59MrRobot (approved):
Good job!

PR #1003, opened 2022-04-27
solution

WojciechCzarnowski:


src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const retobj = {};
```
59MrRobot, line 28: Maybe naming this object something more clear and descriptive?

59MrRobot (approved):
Good job, Wojciech! 👍 just one suggestion though

meforoma (changes requested):
code is good, yet i agree that `retobj` shall be more descriptive, please

src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const inverted = {};
```
VinnieJ-2k20, line 28: That's even better
```suggestion
  const invertedRobot = {};
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #1009, opened 2022-04-27
task is solved

enrikadej:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverseRob = {};
+
+  for (const key in robot) {
+    if (inverseRob.hasOwnProperty(robot[key])) {
+      return null;
+    }
+    inverseRob[robot[key]] = key;
```
anastasiia-nikita, line 33: It would be better to add a blank before for a better visual

Denchepornyuk: agree

Denchepornyuk (approved):

PR #1012, opened 2022-04-27
Repaired robot

Yaroslav-Kononenko:


Anna-Tkachenko (changes requested):
So many iterations...if you have an object with 100 properties, your code should make 100 + 100 + 100 + (100 * 100) + (100 * 100) + 100 iterations (check all iterations in your code)

Let's optimize it, this task can be completed using only one object iteration, only one cycle 🙂

src/inverseRobot.js
```diff
@@ -24,7 +24,36 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.keys(robot);
+  const keys = Object.values(robot);
+  const repairStats = {};
+
+  for (let j = 0; j < values.length; j++) {
+    let sumk = 0;
```
valerii-derkach, line 32: it is not understandable what do the variables mean: sumk, sumv, ch. Use more understandable names

valerii-derkach (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1016, opened 2022-04-28
Solved inverseRobot task

dellachaise:


v-vytrykush (commented):
Good job!

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+
+  for (const value of robotValues) {
+    if (robotValues.indexOf(value) !== robotValues.lastIndexOf(value)) {
```
Denchepornyuk, line 30: use `for in` loop and remove robotValues variable
also, use `hasOwnProperty` instead of `indexOf` and `lastIndexOf`

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+  }
+
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const currentValue = robot[key];
+
+    inversedRobot[currentValue] = key;
+  }
+
```
Denchepornyuk, line 42: you can move these code to the first loop and remove this loop

Denchepornyuk (changes requested):

PR #1018, opened 2022-04-28
inverseRobot

volodya-leskiv:


src/inverseRobot.js
```diff
@@ -23,8 +23,18 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const res = {};
```
Anna-Tkachenko, line 28: What is res? Create more meaningful variable name

Anna-Tkachenko (approved):

PR #1019, opened 2022-04-28
Develop

frontend112:


src/inverseRobot.js
```diff
@@ -24,7 +24,34 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  if (Object.keys(robot).length === 0) {
+    return {};
+  }
```
meforoma, line 29: it's actually much simpler - just declare a blank object variable, say `newRobot = {}`
and you return it from the function - so if no changes will happen it remains a blank object.

And you can fill that object with a `for ( ... in ... )` loop - check if your newRobot hasOwnProperty of an original robot:
- `robot[key]` - should become a key,
- and a `key` - should become a value

meforoma (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobots = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (value.toString().indexOf(' ') > -1 || newRobots.hasOwnProperty(value)) {
```
ihor-jpeg, line 32: I'm not sure what this part does. Let's make it more self explanatory

ihor-jpeg (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+    if (newRobots.hasOwnProperty(value)) {
+      return null;
+    }
+
+    newRobots[value] = key;
+  }
+
+  if (newRobots === {}) {
+    return null;
+  }
```
ihor-jpeg, line 41: in what case will this return true?

ihor-jpeg (changes requested):

meforoma (approved):

PR #1022, opened 2022-04-30
solution

Ihor-Pavlenko:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const ret = {};
```
ArtemTopolskyi, line 27: bad variable name

ArtemTopolskyi (approved):
interesting solution

PR #1032, opened 2022-05-08
solution

vinnibea:


src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+  const newRobot = {};
+  const len = keys.length;
+
+  for (let i = 0; i < len; i++) {
```
illiarerun, line 33: Do you really need index here? You could've used `for ... in` loop

src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+  const newRobot = {};
+  const len = keys.length;
```
illiarerun, line 31: Overkill
```suggestion
```

src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
...
+
+  for (let i = 0; i < len; i++) {
+    if (keys.slice(i + 1, keys.length).includes(keys[i])) {
```
illiarerun, line 34: Same here, that's an overkill, use Object methods to check if object has such property, [Checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 2nd item

src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
...
+      return null;
+    } else {
+      newRobot[keys[i]] = values[i];
+    }
```
illiarerun, line 38: You don't need else here

illiarerun (changes requested):
Nice start, now let's work on some fixes:

src/inverseRobot.js
vinnibea: Got it)

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+  const newRobot = {};
+
+  for (const key in keys) {
```
illiarerun, line 32: use `for ... of` for arrays, that way you'll get a key value, not it's index

illiarerun (changes requested):

PR #1052, opened 2022-05-27
task done

Samaev:


src/inverseRobot.js
```diff
@@ -23,8 +23,53 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  if (!robot) {
+    return {};
+  }
```
illiarerun, line 30: `@param {object} robot`

I guess it's a bit of an overkill, but anyway, GJ

src/inverseRobot.js
```diff
@@ -23,8 +23,53 @@
...
+
+  const keys = Object.keys(robot).reverse();
+  const values = Object.values(robot).reverse();
+  const tem = {};
```
illiarerun, line 34: what is tem? use proper names for variables

src/inverseRobot.js
```diff
@@ -23,8 +23,53 @@
...
  */
+
 function inverseRobot(robot) {
-  // write code here
+  if (!robot) {
+    return {};
+  }
+
+  const keys = Object.keys(robot).reverse();
+  const values = Object.values(robot).reverse();
```
illiarerun, line 33: why do you reverse these?

src/inverseRobot.js
```diff
@@ -23,8 +23,53 @@
...
+      return null;
+    }
+  }
+
+  if (keys.length !== values.length) {
+    return null;
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    tem[values[i]] = keys[i];
```
illiarerun, line 69: That's too much for such a simple task, try to work around `for ... in` loop and go through [checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md), it may help

illiarerun (changes requested):
Nice start, but let's fix your solution:

src/inverseRobot.js
Samaev: now I know about params)
deleted this

src/inverseRobot.js
Samaev: that was temporary in my minds.. changed already

src/inverseRobot.js
Samaev: resulted array with items was reversed,  was thinking that tests won`t pass,
but any way - I found a shorter solution

src/inverseRobot.js
Samaev: OMG, I was young and stu....
now I guess can make shorter

illiarerun (approved):

PR #1057, opened 2022-06-01
Inverse Robot

VolodymyrShymanskyi:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let result = {};
+
+  for (const [key, velue] of Object.entries(robot)) {
+    result[velue] = key;
```
aliceblunt8, line 30:
```suggestion
    if (!result[value]) {
      result[value] = key;
    } else {
      return null;
    }
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+  }
+
+  if (Object.keys(result).length !== Object.keys(robot).length) {
+    result = null;
+  }
```
aliceblunt8, line 35: then you can remove this if-block

aliceblunt8 (changes requested):

VolodymyrShymanskyi:
Changes were made

aliceblunt8 (approved):

PR #1060, opened 2022-06-01
Solution

ik-web:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  for (const key in robot) {
+    if (!robot[robot[key]]) {
```
illiarerun, line 28: [Checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 2nd item, what if value contains falsy value? empty string, null, false, etc. Try to fix your check

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
+      robot[robot[key]] = key;
+      delete robot[key];
+    } else {
+      return null;
+    }
```
illiarerun, line 33: You can invert your condition to avoid using `else` construction

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  for (const key in robot) {
+    if (!robot[robot[key]]) {
+      robot[robot[key]] = key;
+      delete robot[key];
```
illiarerun, line 30: you should never modify function arguments!
And also this action is redundant
```suggestion
```

illiarerun (changes requested):
Nice star, now let's work on fixes:

src/inverseRobot.js
ik-web: Fixed

src/inverseRobot.js
ik-web: Fixed

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty([robot[key]])) {
```

jstmpelowycz (changes requested):

src/inverseRobot.js
ik-web: Fixed

jstmpelowycz (approved):
LGTM!

PR #1063, opened 2022-06-02
solved

mflkvsk:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const robotNew = {};
+
+  for (const key in robot) {
+    if (robotNew[robot[key]]) {
+      return null;
+    }
+
+    if (!robotNew[robot[key]]) {
```
illiarerun, line 34: [Checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 2nd item, what if value contains falsy value? empty string, null, false, etc. Try to fix your check

illiarerun (changes requested):
Nice start, now let's work on one little fix:

src/inverseRobot.js
illiarerun, line 34: And do you really need this check after previous one?

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotNew = {};
+
+  for (const key in robot) {
+    if (robotNew.hasOwnProperty(robot[key])) {
```

illiarerun (approved):

PR #1069, opened 2022-06-05
Solution

ghost:
Solution  Inverse Robot

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverseSettings = {};
+
+  for (const key in robot) {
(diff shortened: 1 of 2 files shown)


vsupruniuk (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
-  // write code here
+  const inverseSettings = {};
+
+  for (const key in robot) {
+    inverseSettings[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(inverseSettings).length) {
+    return null;
+  }
```
jstmpelowycz, line 35: I'd say it's not much obvious, but still an interesting solution

jstmpelowycz (approved):

PR #1072, opened 2022-06-06
Solution

GavrilyukYura:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
illiarerun, line 27: don't use names as a `result` try to come up with something more meaningful

illiarerun (approved):
GJ, left one comment

PR #1073, opened 2022-06-06
Solution

AndriiNychaiuk:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const key in robot) {
+    if (inversed.hasOwnProperty(robot[key])) {
```

jstmpelowycz (changes requested):
awesome! a tiny thing to fix

PR #1083, opened 2022-06-06
Solution

NikitaKarbazin:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (!newRobot.hasOwnProperty(value)) {
+      newRobot[value] = key;
+    } else {
+      return null;
+    }
```
illiarerun, line 34: invert this statements so you wouldn't use `else`
```suggestion
    if (newRobot.hasOwnProperty(value)) {
      return null;
    } 
    
    newRobot[value] = key;
```

illiarerun (approved):
GJ, left one comment:

PR #1090, opened 2022-06-06
Solution

Bohdan-Pyschchyk:


yanaost (approved):
looks reliable

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const value = Object.keys(robot);
+  const key = Object.values(robot);
+  const res = {};
```
illiarerun, line 30: [Checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 3rd item

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+
+  for (let i = 0; i < value.length; i++) {
```
illiarerun, line 32: do you really need index here? can't you use `for ... in` without creating additional variables?

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+    if (res.hasOwnProperty(key[i])) {
+      return null;
+    } else {
+      res[key[i]] = value[i];
+    }
```
illiarerun, line 37: You don't need else here, code will continue to work after this if if condition in falsy anyways
```suggestion
    }

    res[key[i]] = value[i];
```

illiarerun (changes requested):
GJ, now let's work on fixes:

illiarerun (approved):
GJ, but work on your variable names, try to avoid using words that describe variable type

PR #1091, opened 2022-06-06
Solution

IvanGrabovsky:


Samaev (approved):
good job
looks  smart, works great

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
-  // write code here
+  const newObj = {};
+
+  for (const key in robot) {
+    newObj[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(newObj).length) {
+    return null;
+  }
```
illiarerun, line 35: that's okay, but you can use method from [checklist's](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 2nd item, to check properties' presence in obj, and that way return value earlier in loop

illiarerun (approved):
GJ, left one comment:

PR #1093, opened 2022-06-06
Solution

yanaost:


Bohdan-Pyschchyk (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const reversed = {};
+  const values = [];
```
illiarerun, line 28: that array is redundant
```suggestion
```

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const reversed = {};
```
illiarerun, line 27: reversed what? be more specific

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const values = [];
+
+  for (const key in robot) {
+    if (values.indexOf(robot[key]) === -1) {
```
illiarerun, line 31: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 2nd item, there you can see how to properly check if the property is present in object

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+      values.push(robot[key]);
+      reversed[robot[key]] = key;
+    } else {
+      return null;
+    }
```
illiarerun, line 36: Use inverted condition to avoid using null, like:
```
if (condition) {
  return null;
}

regular code here;

illiarerun (changes requested):

PR #1094, opened 2022-06-06
Solution

Kate-Khon:


illiarerun (approved):
Usually I don't approve this task without using `hasOwnProperty` here, but your solution looks interesting, GJ 🔥

PR #1099, opened 2022-06-06
Solution

MariiaNH:


Tsymbal-Viacheslav (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const robotDublicate = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (!(value in robotDublicate)) {
+      robotDublicate[value] = key;
+    } else {
+      return null;
+    }
```
illiarerun, line 36: invert this condition, so you wouldn't need to use `else` here
```suggestion
    if ((value in robotDublicate)) {
      return null;
    }
    
     robotDublicate[value] = key;
```

illiarerun (approved):
GJ, left one comment, pay attention to it:

PR #1113, opened 2022-06-07
Compleated

ant2nta:


dimabelotskyi (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
-  // write code here
+  const newObj = {};
+  const arrayCheck = [];
+
+  for (const prop in robot) {
+    if (!arrayCheck.includes(robot[prop])) {
+      arrayCheck.push(robot[prop]);
+    } else {
+      return null;
+    }
```
jstmpelowycz, line 35:
```
		if (arrayCheck.includes(robot[prop])) {
    	return null;
		}
    
    arrayCheck.push(robot[prop]);
```

jstmpelowycz: we could get it simpler

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+  const arrayCheck = [];
+
+  for (const prop in robot) {
```
jstmpelowycz, line 30: I'd rather call it `key`
I guess, property is a combination of key and value

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
```
jstmpelowycz, line 27: don't be shy when giving your vars names 😉 
```suggestion
  const newObject = {};
```

jstmpelowycz (changes requested):

illiarerun (approved):

PR #1114, opened 2022-06-07
Solution

Dmytrit:


AKomashchuk (approved):
👌

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    const keys = robot[key];
```
illiarerun, line 30: why `keys`? it's a value from base robot, so name it properly, and it shouldn't be plural it's not an array

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+    const values = key;
```
illiarerun, line 31: same here? why plural?
```suggestion
    const value = key;
```

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+    if (newRobot.hasOwnProperty(robot[key])) {
```
illiarerun, line 33: use variable that you've created here

illiarerun (changes requested):
GJ, now let's work on fixes:

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const part in robot) {
+    const key = robot[part];
+    const value = part;
```
jstmpelowycz, line 31: that's redundant, guess you why)

jstmpelowycz (approved):

PR #1125, opened 2022-06-08
solution

Ihor-Novakovskyi:


src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const robot1 = {};
```
illiarerun, line 28: what if you had another robot? would you name it robot2? try to describe what variable is having as a value. `newRobot` would be a great name

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+
+  for (let i = 0; i < Object.values(robot).length; i++) {
+    if (Object.values(robot).slice(i + 1).includes(Object.values(robot)[i])) {
+      return null;
+    }
+  }
+
+  for (const char of Object.entries(robot)) {
```
illiarerun, line 36: why char? that's key-value pair, you can definitely name it `entry`

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const robot1 = {};
+
+  for (let i = 0; i < Object.values(robot).length; i++) {
+    if (Object.values(robot).slice(i + 1).includes(Object.values(robot)[i])) {
```
illiarerun, line 31: this looks overcomplicated, try to shorten it a bit, you can look in [checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 2nd item to find an easier solution

illiarerun (changes requested):
GJ, let's work on fixes now:

illiarerun (changes requested):

mgrinko (approved):

PR #1127, opened 2022-06-10
Solution

dmytro-legerda:


src/inverseRobot.js
```diff
@@ -18,13 +18,25 @@
...
+  const inverseObject = {};
+  const values = Object.values(robot);
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = i + 1; j < values.length; j++) {
+      if (values[i] === values[j]) {
+        return null;
+      }
+    }
+  }
```
vkryvytskyy, line 33: You don't need this, just check if inverseObject has key you're trying to add in loop below and return null if it does

vkryvytskyy (changes requested):

PR #1138, opened 2022-06-21
Solution

muckbet:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(newRobot).length) {
+    return null;
+  }
```
vkryvytskyy, line 35: You don't need this, check if `newRobot` has key you're trying to add in previous loop

vkryvytskyy (changes requested):

PR #1145, opened 2022-06-27
Solution

aleksandra-prognimak:


src/inverseRobot.js
```diff
@@ -24,7 +24,14 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const obj = Object.entries(robot).map(([key, value]) => [value, key]);
```
sergii-nosachenko, line 27: Nice idea of swapping, but constant `obj` is not an regular object, isn't it?) Better name it as `arrayEntries` or even `robotEntries`.

sergii-nosachenko (changes requested):
Hi, interesting approach)
Only found one minor naming issue, not critical.

AndriiRut (approved):
🚀

PR #1148, opened 2022-06-30
the task is done

butenko108:


Ihor-Karpyn (changes requested):
st review required ⚠️

src/inverseRobot.js
```diff
@@ -24,7 +24,28 @@
...
+
+    for (let j = 0; j < keys.length; j++) {
+      if (keys[i] === keys[j]) {
+        count++;
+      }
+    }
+
+    if (count > 1) {
+      return null;
+    }
```
solaryasha, line 43: This code will work, but you don't need to create 2 arrayі, then сreate nested loop for counting duplicates.
You could just check in object that you have duplicated keys (use classic `for in` loop)

solaryasha (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
solaryasha, line 27:
```suggestion
  const inversedRobot = {};
```

solaryasha (approved):

PR #1168, opened 2022-07-03
initial solution

DashaPoliazz:


src/inverseRobot.js
```diff
@@ -25,6 +25,21 @@
...
   // write code here
+  const values = Object.values(robot);
+
+  for (let i = 0; i < values.length; i++) {
+    const value = values[i];
+
+    if (values.indexOf(value) !== values.lastIndexOf(value)) {
+      return null;
+    }
+  }
```
solaryasha, line 36: This code will work, but you don't need to create array, then check.
You could just check in object that you have duplicated keys (use classic `for in` loop)

solaryasha (changes requested):

src/inverseRobot.js
DashaPoliazz: Yap, sorry. I Implemented first solution which was in mind. Yours is better.

solaryasha: It's okay, you could do it with many ways. You had an interesting solition

solaryasha (approved):
Good job !

PR #1181, opened 2022-07-04
Solution

PavloMarkov:


src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  let trueRobot = {};
+
+  for (const key in robot) {
+    if (trueRobot[robot[key]]) {
+      trueRobot = null;
```
illnino380, line 32: [CODE STYLE] - use correct check for property presence in object. Some properties can be present but still contain falsy value.
you chose a bad example with an uncleared solution.

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+
+      return trueRobot;
+    } else {
+      trueRobot[robot[key]] = key;
```
illnino380, line 36: if you return `null` function will be over and no need to add `else`; just put this line after the condition

illnino380 (changes requested):
good job, please make it clear to understand

illnino380 (approved):
Great job!

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  let trueRobot = {};
+
+  for (const key in robot) {
+    if (trueRobot.hasOwnProperty(robot[key])) {
+      trueRobot = null;
+
+      return trueRobot;
```
VETALEK, line 34: you may just `return null` here

VETALEK (approved):
👍

PR #1182, opened 2022-07-04
solution

HusakOleh:


src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const res = { };
```
sergii-nosachenko, line 27:
```suggestion
  const res = {};
```

I think, that this inner space is unnecessary in empty object declaration, do you?

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+    res[robot[key]] = key;
+  }
+
+  const objKeysArr = Object.values(robot).sort();
+
+  for (let i = 0; i < objKeysArr.length; i++) {
+    if (objKeysArr[i] === objKeysArr[i + 1]) {
+      return null;
+    }
+  };
```
sergii-nosachenko, line 39:
```suggestion
```
You don't need this checks at all. You can simply check if the key is already present in res before adding.

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const res = { };
+
+  for (const key in robot) {
+    res[robot[key]] = key;
```
```suggestion
    if (robot[key] in res) {
      return null;
    }
    res[robot[key]] = key;
```
This approach is way more simple and understandable.

sergii-nosachenko (changes requested):
Hi, Oleh!
Your solution is too complicated and can be made more clear and simple. But, of course, you can keep your solution if you want. 🐱‍💻

HusakOleh:
> Hi, Oleh! Your solution is too complicated and can be made more clear and simple. But, of course, you can keep your solution if you want. 🐱‍💻

дякую за допомогу, дійсно, код став більш лаконічним

src/inverseRobot.js
```diff
@@ -24,7 +24,44 @@
...
+    inverse[robot[key]] = key;
+  }
+
+  // const objKeysArr = Object.values(robot).sort();
+  //
+  // for (let i = 0; i < objKeysArr.length; i++) {
+  //   if (objKeysArr[i] === objKeysArr[i + 1]) {
+  //     return null;
+  //   }
+  // };
```
sergii-nosachenko, line 42:
```suggestion
```

Cleanup unnecessary code.

src/inverseRobot.js
```diff
@@ -24,7 +24,44 @@
...
+//   for (let i = 0; i < objKeysArr.length; i++) {
+//     if (objKeysArr[i] === objKeysArr[i + 1]) {
+//       return null;
+//     }
+//   };
+//
+//   return inverse;
+// }
+//
+// module.exports = inverseRobot;
```
sergii-nosachenko, line 67:
```suggestion
```
Cleanup unnecessary code.

src/inverseRobot.js
```diff
@@ -24,7 +24,44 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    if (robot[key] in inverse) {
```
```suggestion
    if (inverse.hasOwnProperty(robot[key])) {
```
This check is better. [Here is why.](https://thisthat.dev/has-own-property-vs-in/)

sergii-nosachenko (changes requested):
Some quick fixes and it will be good to go)

sergii-nosachenko (approved):
Looks perfect! 🔥

MaksimusCRS (approved):

PR #1184, opened 2022-07-04
added inverse Robot function

Oleksandr-Pustovit:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
+
+    if (inversedRobot[newKey]) {
```
yaroslav-lizogub, line 32: checklist 3. use correct check for property presence in object. Some properties can be present, but still contain falsy value.

yaroslav-lizogub (changes requested):

yaroslav-lizogub (approved):
very cool!

VETALEK (approved):
👍

PR #1185, opened 2022-07-04
Solution

LeonidPustovit:


Oleksandr-Pustovit (approved):
It definately works :)

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
-  // write code here
+  const reversedRobot = {};
+  let usedKeys = '';
+
+  for (const key in robot) {
+    reversedRobot[robot[key]] = key;
+
+    if (usedKeys.includes('' + robot[key])) {
+      return null;
+    }
```
VETALEK, line 35: better use `Object.hasOwnProperty` - it makes only one operation (checking the key), but `includes` is looping through the whole string

VETALEK (approved):
Nice 👌

VETALEK (changes requested):
You know, the topis is `objects`, so you need to use object methods in solution 😉

VETALEK (approved):
👍

PR #1197, opened 2022-07-05
Solution

yaroslav-lizogub:


Oleksandr-Pustovit (approved):
Well done!

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const repaired = {};
```
solaryasha, line 27: What was repaired ? It's better to use explicit namings
```suggestion
  const repairedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (robot[key] in repaired) {
+      return null;
+    } else {
+      repaired[robot[key]] = key;
+    }
```
solaryasha, line 34: You don't need `else` keyword here
```suggestion
      repaired[robot[key]] = key;
    }
```

solaryasha (changes requested):

solaryasha (approved):

PR #1202, opened 2022-07-06
Solution

katiailchenko:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
```
ivadimko, line 29: you can safely use `Object.entries` here

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+
+  for (let i = 0; i <= values.length - 1; i++) {
+    if (values[i] in newRobot) {
+      return null;
+    }
+
+    newRobot[values[i]] = keys[i];
```
ivadimko, line 36: here `values[i]` is repeated, would be great to create variables for repeated values
```
const key = values[i]
const value = keys[i]
```

or while using Object.entries:
```
cosnt [value, key] = entry;
```

ivadimko (approved):
a w e s o m e!
🔥 🔥 🔥 🔥

PR #1205, opened 2022-07-06
inverse robot solution

AlexanderKolomiiets:


dellachaise:

Githi54 (approved):
Good job!👍

src/inverseRobot.js
```diff
@@ -23,8 +23,18 @@
...
  */
+
 function inverseRobot(robot) {
-  // write code here
-}
+  const object = {};
 
+  for (const key in robot) {
+    object[robot[key]] = key;
+  }
```
dellachaise, line 32: You could check in your loop is your object already `hasOwnProperty(robot[key])` and return null in this case. This way you code could have less iterations.

src/inverseRobot.js
```diff
@@ -23,8 +23,18 @@
...
+
+  if (Object.keys(object).length < Object.values(robot).length) {
+    return null;
+  } else {
+    return object;
+  }
```
dellachaise, line 38: If you would check copies in you loop(and return null there), then you could always return object here.

dellachaise (changes requested):

dellachaise (approved):
Well done 👍

PR #1206, opened 2022-07-07
solution

Kulchevych:


dellachaise:

sergii-nosachenko (approved):
Hi, Nazar! Perfect solution so far 🔥

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (robot[key] in fixedRobot) {
+      return null;
+    }
```
dellachaise, line 32: Better use `hasOwnProperty` to check if key in object. because `in` will also return true if key gets found somewhere in the prototype chain, otherwise Object.hasOwnProperty (like the name already tells us), will only return true if key is available on that object directly (its "owns" the property).

dellachaise (changes requested):

dellachaise (approved):
Well done 👍

PR #1212, opened 2022-07-10
Solution

KorniychukLesya:


dellachaise:

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+  const result = {};
+  const check = Object.values(robot);
+
+  check.sort();
+
+  for (let i = 0; i < check.length; i++) {
+    if (check[i] === check[i + 1]) {
+      return null;
+    }
+  }
```
sergii-nosachenko, line 36:
```suggestion
```
You don't need this checks at all. You can simply check if the key is already present in `result` before adding.

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+
+  for (const key in robot) {
+    result[robot[key]] = key;
```
sergii-nosachenko, line 39:
```suggestion
    if (robot[key] in result) {
      return null;
    }
    result[robot[key]] = key;
```
This approach is way more simple and understandable.

sergii-nosachenko (changes requested):
Hi, Lesya! ✌

Your solution is too complicated and can be made more clear and simple. But, of course, you can keep your solution if you want.

KorniychukLesya:
Hi, Sergii! Thanks for checking.👍  This is really the best solution.😊

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const check = Object.values(robot);
+
+  check.sort();
```
```suggestion
```

This code is useless. Just remove it without any regret))

sergii-nosachenko (changes requested):

KorniychukLesya:
Thanks a lot!😄 Deleted without regret)) My code has become much shorter)))

sergii-nosachenko (approved):
Perfect! 👍

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (robot[key] in result) {
+      return null;
+    }
```
dellachaise, line 32: better use `hasOwnProperty` to check if key exist in object.

dellachaise (approved):
Well done 👍

src/inverseRobot.js
KorniychukLesya: Thank you! I'll take note😊

PR #1217, opened 2022-07-14
Solution ver 1.0

Oleksii-Korovnichenko:
Hi. 
My solution includes searching the same object values from values array geted from Object.values() method.
But I think we can also use Array.prototype.some(function) with function which also find the same values in object.
Anyway we have a correct result I think. 
Please give me your feedback! Thank you.

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -24,7 +24,32 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  // get array of robot values
+  const robotValues = Object.values(robot);
+  // create new emty object for result
+  const correctRobot = {};
+
+  // searching the same values in values array, if founded then return null
+  for (let i = 0; i < robotValues.length; i++) {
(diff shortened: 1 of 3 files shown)


src/inverseRobot.js
```diff
@@ -24,7 +24,32 @@
...
+  // checking for the no prototype properties and then add correct key values
+  // for new correctRobot Object and then return it
+  for (const prop in robot) {
+    if (robot.hasOwnProperty(prop)) {
+      correctRobot[robot[prop]] = prop;
+    }
+  }
+
+  // console.log(correctRobot);
+  return correctRobot;
```
alexandra-protyanova, line 52: you can just create one `for in` loop
  check is there this key
  replace key & value in places
  and add to new object

alexandra-protyanova (changes requested):
There is a easier decision of this task

valerii-derkach (approved):

PR #1218, opened 2022-07-14
Solution

mariabramnik:


src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
  */
 function inverseRobot(robot) {
   // write code here
+  let inverseNewRobot = {};
+
+  for (const key in robot) {
+    const val = robot[key];
+    const newKey = val;
+
+    if ((newKey in inverseNewRobot) === true) {
```
Bohdan-Maksymiuk, line 34:
```suggestion
    if (newKey in inverseNewRobot) {
```

src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
...
+      inverseNewRobot = null;
+      break;
```
Bohdan-Maksymiuk, line 36:
```suggestion
      return null;
```
you can just return from function here

src/inverseRobot.js
```diff
@@ -25,6 +25,20 @@
  */
 function inverseRobot(robot) {
   // write code here
+  let inverseNewRobot = {};
+
+  for (const key in robot) {
+    const val = robot[key];
```
Bohdan-Maksymiuk, line 31: do not shorten variable name
it is clear what it means, but take it into consideration for the next tasks

Bohdan-Maksymiuk (changes requested):

src/inverseRobot.js
mariabramnik: I try not to do more than one return

Bohdan-Maksymiuk: it is not a problem when you want to end function

Bohdan-Maksymiuk (approved):

Bohdan-Maksymiuk (approved):

PR #1227, opened 2022-07-22
task done

sergey-vorobei:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (!result[robot[key]]) {
+      result[robot[key]] = key;
+    } else {
+      return null;
+    }
```
Bohdan-Maksymiuk, line 34:
```suggestion
    if (result[robot[key]]) {
      return null;
    } 
    
    result[robot[key]] = key;
```

Bohdan-Maksymiuk: Also better to use hasOwnProperty for checking if keys exist in the object

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1231, opened 2022-07-24
Solution

serhii-yunnikov:


src/inverseRobot.js
```diff
@@ -24,7 +24,27 @@
...
+  const values = Object.values(robot);
+  const result = {};
+
+  function matches(arr) {
+    for (const val of arr) {
+      for (const val1 of arr) {
+        return val === val1 && arr.indexOf(val) !== arr.lastIndexOf(val1);
+      }
+    }
+  }
```
Bohdan-Maksymiuk, line 37: try don't use nested loops
it increases complexity of your code

Bohdan-Maksymiuk (changes requested):
you can use only 1 loop and iterate through keys of robot - `for (const key in robot)`
and inside check `if (newRobot.hasOwnProperty(robot[key]))`
you need to write what to do depend on this condition

Bohdan-Maksymiuk (approved):

PR #1233, opened 2022-07-25
solution

AndreyJigijalkin:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newBot = {};
+
+  for (const keys in robot) {
+    newBot[robot[keys]] = keys;
+  }
```
AlexanderKolomiiets, line 31: Everything is great, but only one little comment which is really super unimportant but still. )
You better name const keys in singular (just key) because it contains only one value.

AlexanderKolomiiets (approved):
Great !

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+
+  if (Object.keys(newBot).length
+  !== Object.keys(robot).length) {
```
MaksimusCRS, line 34: Why do you need such logic? Just use hasOwnProperty

MaksimusCRS (approved):

PR #1236, opened 2022-07-26
all done

Yevhen-Pavlovskyi:


src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverse = {};
+  const values = Object.values(robot);
+
+  for (let i = 0; i < values.length; i++) {
```
alexandra-protyanova, line 30: u can just use for in loop for robot & and do all u need instead of creating 3 loops

alexandra-protyanova (changes requested):

valerii-derkach (approved):

PR #1245, opened 2022-08-02
Solution

Dmytro-Matsehora:


dellachaise:

Kulchevych (approved):
Good job

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    fixedRobot[robot[key]] = key;
```
dellachaise, line 30: Better check in for cycle is `key` already exist in `fixedRobot` and return `null` for non unique. This way you could rewrite 33-36 lines too. And it will end program with `null` quicker without unnecessary operations.

dellachaise (changes requested):

src/inverseRobot.js
Dmytro-Matsehora: thanks for your advice, it really reduced the amount of code made it cleaner 👍

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (robot[key] in fixedRobot) {
+      return null;
+    }
```
dellachaise, line 32: Better use `hasOwnProperty` to check is property in object. because `in` will also return true if key gets found somewhere in the prototype chain, otherwise Object.hasOwnProperty (like the name already tells us), will only return true if key is available on that object directly (its "owns" the property).

dellachaise (approved):
Well done 👍 

But rewrite with `hasOwnProperty`.

PR #1253, opened 2022-08-03
Develop

kir8mir:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
-  // write code here
+  const returnedObject = {};
+
+  for (const key in robot) {
+    if (returnedObject.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      returnedObject[robot[key]] = key;
+    }
+  }
```
tiserett, line 35:
```suggestion
  for (const key in robot) {
    if (returnedObject.hasOwnProperty(robot[key])) {
      return null;
    }

    returnedObject[robot[key]] = key;
  }
```
no need to add else statement

tiserett (approved):

src/inverseRobot.js
AndriiRut: it's true!)

AndriiRut (approved):

PR #1254, opened 2022-08-03
Solution

nikitaovch99:


artem-musii (approved):
You can use if statement instead of second loop, but it's up to you

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+  const rb1 = Object.keys(robot);
+  const rb2 = Object.values(robot);
+  const obj = {};
+
+  for (let i = 0; i < rb1.length; i++) {
+    for (let g = i + 1; g < rb1.length; g++) {
+      if (rb1[i] === rb1[g] || rb2[i] === rb2[g]) {
+        return null;
+      }
+    }
```
naziamoff, line 36: You definitely have to avoid nested loops in your code

naziamoff: Btw your variable names are unclear. You should think of sth that helps to understand what is stored inside the variable

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const rb1 = Object.keys(robot);
+  const rb2 = Object.values(robot);
```
naziamoff, line 28: Actually you don't need those to complete the task. There is a simpler way

Try to do it with help of `for...in` loop and  `.hasOwnProperty()` method

naziamoff (changes requested):

naziamoff (approved):
Nicely done 🚀 
I approve but it is recommended to use full naming in future, check comments

PR #1260, opened 2022-08-04
Solution

BipBobBoop:


src/inverseRobot.js
```diff
@@ -24,7 +24,13 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    copy[value] = key;
+  }
+
+  return Object.keys(robot).length === Object.keys(copy).length ? copy : null;
```
Anna-Tkachenko, line 33: For now you used 4 iterations to complete this task
1 - for cycle
2 - Object.entries (1)
3, 4 - Object.keys (2 times)

Try to rewrite the solution using only one iteration

Just go through the initial robot keys, and add the needed values to the new object
And add some check inside the cycle to check, if this property is already exists in the new robot

Anna-Tkachenko (changes requested):

Bohdan-Maksymiuk (approved):

PR #1261, opened 2022-08-04
Solution

oksana-logos-frontend:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const swapped = {};
+
+  for (const key in robot) {
+    if (swapped[robot[key]]) {
```
```suggestion
    if (swapped.hasOwnProperty(robot[key])) {
```
Better use own property check for object. [Here is described why](https://masteringjs.io/tutorials/fundamentals/hasownproperty). Of course, it always depends on situation. So its on your own decision.

sergii-nosachenko (approved):
Nice work! 👍
Only have one recommendation to improve.

src/inverseRobot.js
dellachaise, line 30: Better use `hasOwnProperty` to check if property in object. because in your case will also return true if key gets found somewhere in the prototype chain, otherwise Object.hasOwnProperty (like the name already tells us), will only return true if key is available on that object directly (its "owns" property).

dellachaise (changes requested):
Great job 👍 

But please rewrite with `hasOwnProperty`.

src/inverseRobot.js
oksana-logos-frontend: Thanks!Agree)))I'll rewrite it)

dellachaise (approved):
Well done 👍

PR #1262, opened 2022-08-05
Solution

zhenija:


src/inverseRobot.js
```diff
@@ -24,7 +24,26 @@
...
+  let parameter = '';
+  const keys = Object.keys(robot);
+
+  for (let i = 0; i < keys.length; i++) {
+    for (let j = i + 1; j < keys.length; j++) {
+      if (robot[keys[i]] === robot[keys[j]]) {
+        return null;
+      }
+    }
+  }
```
Bohdan-Maksymiuk, line 38: try to avoid such nested loops

Bohdan-Maksymiuk (changes requested):
you can do it into 1 loop 
```
  const newRobot = {};

  for (const key in robot) {
      return null;
    }
    newRobot[robot[key]] = key;
  }

  return newRobot;
```

src/inverseRobot.js
zhenija: Looks more simple, thanks!

Bohdan-Maksymiuk (approved):

PR #1264, opened 2022-08-05
Solution

Ihor-Zakharov:


src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
...
+
+  const values = Object.values(robot);
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = 0; j < values.length; j++) {
+      if (i !== j && values[i] === values[j]) {
+        return null;
+      }
+    }
+  }
```
alexandra-protyanova, line 37: it is not the best solution, u can use check in your `for in` loop

alexandra-protyanova (changes requested):

SanyaBratashchuk (approved):

PR #1275, opened 2022-08-07
solutionInverseRobot

YuriiPronin:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      result[robot[key]] = key;
+    }
```
tiserett, line 34:
```suggestion
    if (result.hasOwnProperty(robot[key])) {
      return null;
    } 
    
    result[robot[key]] = key;
```

it's not neccessary to use 'else' statement, because 'if' statement returns value, which means it will prevent all other code from executing. Not sure about code rules, but I feel like this version of code looks cleaner.

tiserett (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
+      return null;
+    }
+    result[robot[key]] = key;
```
AndriiRut, line 33:
```suggestion
    }
    
    result[robot[key]] = key;
```

AndriiRut (changes requested):

AndriiRut (approved):

PR #1276, opened 2022-08-07
inverseRobot was added

Andrii-Stiebalov:


artem-kodesnikov (approved):
i think better to use method hasOwnProperty for check properties, but your way probably also ok

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+  const newObj = {};
+
+  for (let i = 0; i < values.length; i++) {
+    newObj[values[i]] = keys[i];
+
+    if (keys.indexOf(keys[i]) !== i || values.indexOf(values[i]) !== i) {
```
AndriiRut, line 34: `hasOwnProperty` will be better in use

AndriiRut (changes requested):

AndriiRut (approved):
🔥

PR #1287, opened 2022-08-08
ReparedRobot

Reptiloid044:
Hi!

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const ReparedRobot = {};
+
+  for (const key in robot) {
+    if (ReparedRobot[robot[key]]) {
+      return null;
+    }
+    ReparedRobot[robot[key]] = key;
+  }
+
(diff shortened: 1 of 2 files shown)


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const ReparedRobot = {};
+
+  for (const key in robot) {
+    if (ReparedRobot[robot[key]]) {
```
plemiannik, line 30: better 2 check 4 null before creating new robot

plemiannik (approved):
nice job 👍

src/inverseRobot.js
danylo-onishchuk: agree

danylo-onishchuk: Better use hasOwnProperty to check is property in object. because in will also return true if key gets found somewhere in the prototype chain, otherwise Object.hasOwnProperty (like the name already tells us), will only return true if key is available on that object directly (its "owns" the property).

danylo-onishchuk (approved):

PR #1289, opened 2022-08-08
Solution

Anastasiia-Khmelovska:


max-kravchenko (approved):
Great!! 
p.s. I would only change the name of a variable (inversRobot). I got the logic behind naming, but maybe a different name would fit better.

danylo-onishchuk (approved):

PR #1295, opened 2022-08-08
task complette

Sasha-Gordijchuk:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    inverse[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(inverse).length) {
+    return null;
+  }
```
plemiannik, line 35: fire

plemiannik (approved):
interesting solution with length of keys arrays, but minus is that the check is after instead of before creating those, what is no need, if check fiind, what it search

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    inverse[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(inverse).length) {
```
MaksimusCRS, line 33: Why do you need such logic? Just use hasOwnProperty

MaksimusCRS (changes requested):

MaksimusCRS (approved):

PR #1299, opened 2022-08-08
solution

roma-shafihullin:


src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const reverse = {};
+
+  for (const key in robot) {
+    if (reverse.hasOwnProperty(robot[key])) {
```
plemiannik, line 30: what reason in reverse if robot have 2+ same values

plemiannik (approved):
good work👍

src/inverseRobot.js
AndriiRut: I think the naming of const is confused you) Everything looks good)

but in future @roma-shafihullin  try to use more semantic names because arrays has such method named as reverse)

AndriiRut (approved):

PR #1303, opened 2022-08-09
solution

David-Shahmatov:


src/inverseRobot.js
```diff
@@ -24,7 +24,26 @@
...
+    arr.push(robot[key]);
+  };
+
+  const newArr = arr.sort();
+
+  for (let i = 0; i < newArr.length - 1; i++) {
+    if (newArr[i] === newArr[i + 1]) {
+      return null;
+    };
+  };
```
AndriiRut, line 40: you can use `hasOwnProperty` instead of code from 30 to 40 line
image

AndriiRut (changes requested):

dellachaise (approved):
Well done 👍

PR #1311, opened 2022-08-09
solution

vadym-zahazei:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const repairedRobot = {};
+  const robotData = Object.entries(robot);
+
+  for (const data of robotData) {
+    const key = data[1];
+    const velue = data[0];
```
AndriiRut, line 32: use more semantic naming.

AndriiRut: or rewrite without using 1 and 0 indexes

AndriiRut (changes requested):

src/inverseRobot.js
vadym-zahazei: > use more semantic naming.

done

AndriiRut (approved):
🔥

PR #1331, opened 2022-08-17
Develop

offsset:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const masRobot = Object.values(robot);
+
+  for (const keyMR of masRobot) {
+    for (const key in robot) {
+      if (keyMR === robot[key]) {
+        if (result[robot[key]]) {
+          return null;
```
marina-tilniak, line 34: use for in and hasOwnProperty here

marina-tilniak: you don't need extra array

marina-tilniak (changes requested):

SanyaBratashchuk (approved):

PR #1342, opened 2022-08-25
Solution

AMahalias:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const arr = Object.entries(robot);
+  const values = Object.values(robot).sort();
+
+  for (let i = 0; i < values.length; i++) {
```
alexandra-protyanova, line 31: u can iterate through the object using loop `for in`

alexandra-protyanova (changes requested):

stanislav-topikha (approved):

PR #1363, opened 2022-09-05
Solution

emxm:


src/inverseRobot.js
```diff
@@ -23,8 +23,19 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (robot[key] in result) {
```
Bohdan-Maksymiuk, line 31: better to use `hasOwnProperty()`

src/inverseRobot.js
```diff
@@ -23,8 +23,19 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
Bohdan-Maksymiuk, line 28:
```suggestion
  const inversedRobot = {};
```
use clear naming for variables

src/inverseRobot.js
```diff
@@ -23,8 +23,19 @@
...
+
+  for (const key in robot) {
+    if (robot[key] in result) {
+      return null;
+    } else {
```
Bohdan-Maksymiuk, line 33: and you can get rid of this `else`

Bohdan-Maksymiuk (changes requested):

src/inverseRobot.js
emxm: I ridded "else", but the code became got longer, though more obvious

src/inverseRobot.js
```diff
@@ -23,8 +23,21 @@
...
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (inversedRobot.hasOwnProperty(robot[key]) === true) {
+      return null;
+    }
+
+    if (inversedRobot.hasOwnProperty(robot[key]) === false) {
+      inversedRobot[robot[key]] = key;
+    }
```
Bohdan-Maksymiuk, line 37:
```suggestion
    if (inversedRobot.hasOwnProperty(robot[key])) {
      return null;
    }

    inversedRobot[robot[key]] = key;
```

Bohdan-Maksymiuk (changes requested):

valerii-derkach (approved):

PR #1365, opened 2022-09-05
Solution

MaxManis:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const res = {};
```
Anna-Tkachenko, line 27: what is res? use clear and meaningful variable names

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+
+  for (const n in robot) {
```
Anna-Tkachenko, line 29: what is n?

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #1369, opened 2022-09-06
Solution

devWizHub:


src/inverseRobot.js
```diff
@@ -25,6 +25,26 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  const keys = Object.keys(robot);
+
+  for (let i = 0; i < keys.length; i++) {
```
amitpatiljc, line 32: instead of 2 loops see if you can do it without loops 
may be with set
check this resource
[Link](https://javascript.plainenglish.io/storing-unique-values-with-set-object-in-javascript-ac2707dbdb6)

src/inverseRobot.js
```diff
@@ -25,6 +25,26 @@
...
+    for (let j = i + 1; j < keys.length; j++) {
+      if (robot[keys[i]] === robot[keys[j]]) {
+        return null;
+      }
+    }
+  }
+
+  Object.entries(robot).forEach(entry => {
```
amitpatiljc, line 40: there is better way

amitpatiljc (changes requested):

src/inverseRobot.js
devWizHub: implemented. thanks.

src/inverseRobot.js
devWizHub: Please provide hint. thanks

src/inverseRobot.js
```diff
@@ -25,6 +25,24 @@
...
+    const key = entry[1];
+    const value = entry[0];
+
+    newRobot[key] = value;
+  });
+
+  return newRobot;
```
amitpatiljc, line 45: use `for in` loop and iterate through object
reverse  key and value by
`newObject[OldObject[key]] = key `

amitpatiljc (changes requested):

src/inverseRobot.js
```diff
@@ -25,6 +25,28 @@
...
+
+  const keys = Object.values(robot);
+
+  const newKeys = [...new Set(keys)];
+
+  if (newKeys.length !== keys.length) {
+    return null;
+  }
+
+  for (const k in robot) {
```
amitpatiljc, line 38: use proper names instead of random characters

src/inverseRobot.js
```diff
@@ -25,6 +25,28 @@
...
+    newRobot[robot[k]] = k;
+  }
+
+  // Object.entries(robot).forEach(entry => {
+  //   const key = entry[1];
+  //   const value = entry[0];
+
+  //   newRobot[key] = value;
+  // });
```
amitpatiljc, line 47: delete commented lines of code

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job!

PR #1375, opened 2022-09-10
solution was added

Maksym845:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const prev = [];
+  const res = {};
+
+  for (const ch in robot) {
```
Andr1yK, line 30: what is ch?

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const prev = [];
```
Andr1yK, line 27: redundant, u just increase complexity :c

Andr1yK (changes requested):

Andr1yK (changes requested):
U change nothing 🥲

Now the complexity of this code only increased :(

src/inverseRobot.js
```diff
@@ -24,7 +24,16 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const res = {};
```
Andr1yK, line 27:
```suggestion
  const inversedRobot = {};
```

Andr1yK (approved):

PR #1376, opened 2022-09-10
solution 1.0

aliliolek:


stanislav-topikha (changes requested):
lets do it in one loop, not three)

src/inverseRobot.js
```diff
@@ -14,17 +14,30 @@
...
+ * inverseRobot(kolli) === { name: 'Kolli', chipVer: '123', wheels: '3' }
+ * inverseRobot(robert) === null
  *
  * @param {object} robot
  *
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
ArturErmolenko, line 30:
```suggestion
  const newRobot = {};
```

src/inverseRobot.js
```diff
@@ -14,17 +14,30 @@
...
+
+  for (const key in robot) {
+    result[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(result).length) {
+    return null;
+  }
```
ArturErmolenko, line 38:
```suggestion
  for (const key in robot) {
    result[robot[key]] = key;
  }

  if (Object.keys(robot).length !== Object.keys(result).length) {
    return null;
  }
```
use hasOwnProperty
```
    ```
 const value = robot[key];

     if (newRobot.hasOwnProperty(value)) {
       return null;
     }
```
```

**Review by ArturErmolenko — CHANGES_REQUESTED** [review]

File: src/inverseRobot.js (hunk 2 above)
Comment (aliliolek) [reply src/inverseRobot.js:38]: I haven't understood anything

Comment (aliliolek) [reply src/inverseRobot.js:38]: I've understood nothing

File: src/inverseRobot.js [hunk 3]
```diff
@@ -14,17 +14,34 @@
...
+      return null;
+    }
+
+    newRobot[robot[key]] = key;
+  }
+
+  // if (Object.keys(robot).length !== Object.keys(newRobot).length) {
+  //   return null;
+  // }
+
```
liliia-livandovskaya, line 43:
```suggestion


```
remove comments

liliia-livandovskaya (approved):
GJ 👍

PR #1378, opened 2022-09-11
solution

artem-musii:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
```
naziamoff, line 27:
```suggestion
  const inversedRobot = {};
```

naziamoff: better to use clear naming here. We want to have our code maximally clear

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+
+  for (const key in robot) {
+    if (!inversed[robot[key]]) {
```
naziamoff, line 30: Try `.hasOwnProperty` method here

naziamoff (changes requested):
Well done 🔥 , but there are some minor mistakes

naziamoff (approved):
Gj 🔥

PR #1380, opened 2022-09-12
Solution

vladimir-dublya:


myroslav-diiak (approved):

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const res = {};
+
+  for (const key in robot) {
+    if (res.hasOwnProperty(robot[key])) {
```
Denchepornyuk, line 31: create variable for `robot[key]` inside the loop

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const res = {};
```
Denchepornyuk, line 28: res isn't a good name for variable

Denchepornyuk (changes requested):

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const result = {};
+
+  for (const key in robot) {
+    const Key = robot[key];
```
Denchepornyuk, line 31: why `Key`?
I see you have `key` in the loop so create another name for this variable without uppercase

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+
+    if (result.hasOwnProperty(robot[key])) {
```
Denchepornyuk, line 33: use variable here instead of `robot[key]`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1383, opened 2022-09-12
Solution

AntonVaida:


imaksweb (approved):
GJ! Also you can try to refactor your code using `for in` and `hasOwnProperty`

Andr1yK (changes requested):
It's original, but not a good solution😅

Complexity: n^2 + n
But u can solve this with linear complexity

Andr1yK (approved):

PR #1387, opened 2022-09-12
solution

volodymyrklymkiv:


src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const obj = {};
```
vmamchur, line 28: i'm not sure if obj is a good name 🦧

vmamchur (approved):

src/inverseRobot.js
naziamoff: Vadym is right, this name should be changed
Your variables have to really give some information about what is stored inside. That helps you and other developers understand your code

naziamoff (changes requested):

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const inverse = {};
```
naziamoff, line 28: I think this one looks better
```suggestion
  const inversedRobot = {};
```

naziamoff (approved):
Well done 🚀 
I approve, but it is better to use full and clear namings in future

PR #1389, opened 2022-09-12
Solution

glugovsky:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let newRobot = {};
+
+  for (const key in robot) {
+    if (!newRobot[robot[key]]) {
```
Anna-Borodaienko, line 30: ```(!newRobot[robot[key]])```
Checklist doesn't recommend this, we have to use hasOwnProperty method here

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+      newRobot[robot[key]] = key;
```
Anna-Borodaienko, line 31: You can create a variable robot[key], if you use it more than in 1 place

Vladosik1337 (approved):
GJ)

Denchepornyuk (approved):

PR #1390, opened 2022-09-12
Solution

Apollonia5:


src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverseObj = {};
+
+  for (const objValue of Object.values(robot)) {
```
Yevheniya-Deryka, line 29: Seems to me it'll be better to use `const key in robot` here

Yevheniya-Deryka: Your code will be much more shorter in that case

src/inverseRobot.js
```diff
@@ -24,7 +24,21 @@
...
+    if (inverseObj[objValue]) {
+      return null;
+    }
+
+    for (const objKey of Object.keys(robot)) {
```
Yevheniya-Deryka, line 34: With `const key in robot` there'll be no need in this second loop

Yevheniya-Deryka (approved):
Left some comments. As to me solution with key in looks more neat. But maybe your variant also can exist)

Andr1yK (changes requested):
U need 2 `for` loops?)

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const inverseObj = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (inverseObj[value]) {
+      return null;
+    } else {
+      inverseObj[value] = key;
+    }
```
Andr1yK, line 36:
```suggestion
    if (inverseObj[value]) {
      return null;
    }
      inverseObj[value] = key;
```

Andr1yK (approved):

PR #1393, opened 2022-09-12
Solution

jojnajbude:


src/inverseRobot.js
```diff
@@ -24,7 +24,26 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+
+  for (let i = 0; i < keys.length; i++) {
+    if (keys.slice(i + 1).includes(keys[i])) {
+      return null;
+    }
```
ArturNagaycev, line 33: you don't need to do it. looking at example we must to check only future keys, and not values

 * Example:
 *
 * const kolli = { Kolli: 'name', 123: 'chipVer', 3: 'wheels' };
 * const robert = { Robert: 'name', 123: 'chipVer', 113: 'chipVer' };
 * inverseRobot(robert) === null
 * inverseRobot(kolli) === { name: 'Kolli', chipVer: '123', wheels: '3' }

ArturNagaycev (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+    if (values.slice(i + 1).includes(values[i])) {
+      return null;
+    }
+  }
+
+  const inverse = {};
+
+  for (let i = 0; i < values.length; i++) {
+    inverse[values[i]] = keys[i];
+  }
```
OleksiiBlezniuk, line 40: Do you really need two loops here? Also `slice` and `includes` used loops too
You can just use one `for ... in` loop and read from object by key to check if this property is already in object

OleksiiBlezniuk (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+
+  const inverse = {};
+
+  for (let i = 0; i < values.length; i++) {
+    if (values.slice(i + 1).includes(values[i])) {
+      return null;
+    }
+    inverse[values[i]] = keys[i];
```
OleksiiBlezniuk, line 36: Ok, but still better just to iterate object, no need to create two arrays

OleksiiBlezniuk (approved):

PR #1394, opened 2022-09-12
Task was solved

DenysKleimenov:


Yuriy-Komolov:
Seems perfect to me. Well done!

Yuriy-Komolov (approved):

naziamoff (approved):
Great job! 🔥 
But I would not use name rightRobot, I would better come up with sth like inversedRobot

PR #1395, opened 2022-09-12
Solution

Yevheniya-Deryka:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const invertRobot = {};
+
+  for (const key in robot) {
+    if (invertRobot.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      invertRobot[robot[key]] = key;
```
Anna-Borodaienko, line 33: I'm not sure, that it's necessary, but you can create a variable 'robot[key]' and use it

Yevheniya-Deryka: I agree, it can make sense

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const invertRobot = {};
+
+  for (const key in robot) {
+    const property = robot[key];
+
+    if (invertRobot.hasOwnProperty(property)) {
+      return null;
+    } else {
+      invertRobot[property] = key;
+    }
```
Denchepornyuk, line 36: you can remove else block
```suggestion
    if (!invertRobot.hasOwnProperty(property)) {
      invertRobot[property] = key;
    }
```

Denchepornyuk (approved):

PR #1396, opened 2022-09-12
solution

artem-nasadchuk:


anton-iskryk (changes requested):

anton-iskryk (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (robot[key] in inversedRobot) {
```
naziamoff, line 30: would be better to use .hasOwnProperty method

naziamoff (approved):
Approved, good job 🔥 
But I strongly recommend you to redo it with hasOwnProperty method, check comments

PR #1399, opened 2022-09-12
Solution

iPanda177:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newCopy = {};
```
iam-datsenko, line 27: Copy is new by default, maybe better to call copy or copyRobot

iam-datsenko (approved):

Denchepornyuk (approved):

PR #1400, opened 2022-09-12
Inverse Robot

Vladosik1337:


glu3s (approved):
GJ!

src/inverseRobot.js
```diff
@@ -23,8 +23,21 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  const inverseObj = {};
```
naziamoff, line 28: You should make it maximally clear. Let's say 
```suggestion
  const inversedRobot = {};
```

naziamoff (approved):
Well done! 🔥 
But check my comment about naming

PR #1403, opened 2022-09-12
Solution

Rina07:


src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (invertedRobot[robot[key]]) {
```
Oleksandra-Shevchenko, line 31: i think better to do variables for if ' robot[key] ', and use .hasOwnProperty for checking :)

Oleksandra-Shevchenko (approved):

Anna-Tkachenko (approved):

PR #1406, opened 2022-09-12
Solution

SanyaKovalenko:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const object = {};
```
naziamoff, line 27: it would be better to use more clear naming here.
```suggestion
  const fixedRobot = {};
```

naziamoff (approved):
Nice 🔥 
But check comments, in future you should use full and maximally clear naming

PR #1407, opened 2022-09-12
Solution

rstarovoit:


MAKSOti (approved):
this works but loop within loop is too much work this can be done much easier

src/inverseRobot.js
```diff
@@ -24,7 +24,37 @@
...
+//     }
+//   }
+
+//   for (const i in keys) {
+//     obj[values[i]] = keys[i];
+//   }
+
+//   return obj;
+// }
+
```
Andr1yK, line 59: remove needless comments

Andr1yK (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,37 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
+
+  for (const i in robot) {
+    if (obj.hasOwnProperty(robot[i])) {
+      return null;
+    }
+
+    obj[robot[i]] = i;
```
Andr1yK, line 34: u can use `for of`

Andr1yK (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
+
+  for (const i in robot) {
```
silchyn, line 29: What means "obj"? Also, "i" is for iterator, here it is key

silchyn (changes requested):

src/inverseRobot.js
rstarovoit: "obj" means "object'" from the task _"...and returns a new object in which..."_ but I agree "new Robot" is more understandable) Thanks for "i"!

silchyn (approved):
Cool!

PR #1408, opened 2022-09-12
Solution

OleksandrGlulhovskiy:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
+
+  for (const key in robot) {
+    if (!reverseRobot[robot[key]]) {
```
Andrew-Skoryk, line 30: exp
According to the checklist, it's not the best solution, sadly...

Andrew-Skoryk (changes requested):
I would recommend  to fix "IF" check

Denchepornyuk (approved):

PR #1411, opened 2022-09-12
add decision for task

SerhiiRomaniukWD:


src/inverseRobot.js
```diff
@@ -23,8 +23,25 @@
...
+
+  for (const item of organs) {
+    if (organs.filter(organ => organ === item).length > 1) {
+      return null;
+    }
+
+    return reverseRobot;
+  }
+
+  return inputRobot;
```
Anna-Borodaienko, line 44: Looks like last return is never happened. 
Why do you use it? We don't need to return input robot according to task conditions

SerhiiRomaniukWD: No, whitout this return everything is falling apart. 
When everything okey with robot, we must return clear input

src/inverseRobot.js
```diff
@@ -23,8 +23,25 @@
...
  * @return {object}
  */
-function inverseRobot(robot) {
-  // write code here
+function inverseRobot(robot = {}) {
+  const inputRobot = { ...robot };
+  const reverseRobot = {};
+  const organs = [];
+
+  for (const property of Object.entries(inputRobot)) {
```
Andr1yK, line 31:
```suggestion
  for (const [value, key] of Object.entries(inputRobot)) {
```

Andr1yK (approved):

src/inverseRobot.js
SerhiiRomaniukWD: okey, i will fix it

PR #1414, opened 2022-09-12
solution

La-La-Op:


VovaDzhuryk:
Good job!

VovaDzhuryk (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+
+  for (const key in robot) {
+    if (newObj.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    newObj[robot[key]] = key;
```
Denchepornyuk, line 34: let's create variable for `robot[key]`
you use this value twice in the code

Denchepornyuk (changes requested):

src/inverseRobot.js
La-La-Op: done

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
```
OleksiiBlezniuk, line 27: Create more meaningful variables names, what is stored in this newObj?

OleksiiBlezniuk, line 27:
```suggestion
  const inversedRobot = {};
```

OleksiiBlezniuk (approved):

PR #1416, opened 2022-09-12
implemented task

illia-sosnytskyi:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (inversedRobot.hasOwnProperty(value)) {
```
Andr1yK, line 32: Something wrong with naming, cuz hasOwnProperty checks `keys` but not the values
```suggestion
    if (inversedRobot.hasOwnProperty(value)) {
```

Andr1yK (changes requested):

Andr1yK (approved):

PR #1417, opened 2022-09-12
Solution

Dikey945:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const key = Object.values(robot);
+  const val = Object.keys(robot);
```
illia-sosnytskyi, line 28: it's better to call arrays in the prural form

illia-sosnytskyi (commented):
GJ !

src/inverseRobot.js
illia-sosnytskyi, line 28: it's better to call arrays in the prural form 
```suggestion
  const vals = Object.values(robot);
  const keys = Object.keys(robot);
```

Dikey945: About plural form agree, but about meaning, in this task, we must put reverse: keys > values, values > keys;

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.values(robot);
+  const vals = Object.keys(robot);
```
naziamoff, line 28: Would be better to name variables more clear

```suggestion
  const robotKeys = Object.values(robot);
  const robotValues = Object.keys(robot);
```

naziamoff (approved):
GJ 🔥 
I approve but fix naming pls

src/inverseRobot.js
Dikey945: > Would be better to name variables more clear
Done!)

PR #1418, opened 2022-09-12
Robot's inversion

Yuriy-Komolov:


DenysKleimenov (approved):
GJ!

src/inverseRobot.js
```diff
@@ -25,6 +25,16 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+
+  for (const [value, key] of Object.entries(robot)) {
+    if (key in inversedRobot) {
```
naziamoff, line 31: it is better to use .hasOwnProperty method in that case

naziamoff (approved):
Well done 🔥 
But don't forget about `.hasOwnProperty` method

PR #1419, opened 2022-09-13
complete the task

barsuchellos:


ArtemHulkoDev (approved):
everything is fine)

src/inverseRobot.js
```diff
@@ -24,7 +24,28 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const val = Object.values(robot);
+
+  for (let counter = 0; counter <= val.length; counter++) {
+    let c = 1;
+
+    for (; c < val.length; c++) {
```
Denchepornyuk, line 32: you don't need two loops, you need to create a new empty object, iterate over the keys from the robot object and write the robot value as the key to the new object

src/inverseRobot.js
```diff
@@ -24,7 +24,28 @@
...
+      if (val[counter] === val[c]) {
+        // eslint-disable-next-line no-param-reassign
+        robot = null;
```
Denchepornyuk, line 35: nope, you can rewrite function arguments

denys-cheporniuk (changes requested):

src/inverseRobot.js
barsuchellos: ok , I rewrite it)

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
+
+  for (const property in robot) {
+    if (obj.hasOwnProperty(robot[property])) {
+      return null;
+    }
+
+    obj[robot[`${property}`]] = property;
```
Denchepornyuk, line 34: create variable for `robot[property]` and use a variable in two places

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

Andr1yK (approved):

PR #1429, opened 2022-09-13
Solution

vovchuk-nazar:


MAKSOti (approved):
GJ)) 
but i would use other variable names 
"result" it's not very clear what it is

Andr1yK (approved):

PR #1434, opened 2022-09-13
result

AndriyZhabynets:


Arsen-Shpak (approved):

src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
```
Andr1yK, line 29: use [Object.entries()](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)

src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
...
+
+  for (let i = 0; i < keys.length; i++) {
+    newObj[`${values[i]}`] = keys[i];
+  }
+
+  const ordered = {};
+
+  Object.keys(newObj).sort().forEach(function(key) {
```
Andr1yK, line 37: Do u need this sort?

src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
...
+    ordered[key] = newObj[key];
```
Andr1yK, line 38: Mb, can we check if this field already exist in new object then return null?

Andr1yK (changes requested):

src/inverseRobot.js
AndriyZhabynets: ok

src/inverseRobot.js
AndriyZhabynets: ok

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (newObj.hasOwnProperty(value)) {
+      return null;
+    }
+
+    newObj[`${value}`] = key;
```
Andr1yK, line 34:
```suggestion
    newObj[value] = key;
```

Andr1yK (approved):

PR #1436, opened 2022-09-14
Solution

vaadyak:


Andr1yK (changes requested):
Where are students' reviews?

MAKSOti (approved):
GJ))
but i would use other variable names
"result" it's not very clear what it is

Andr1yK (approved):

PR #1442, opened 2022-09-14
Solution

AleksanderPl:


src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const x = {};
```
marina-tilniak, line 28: use meaningful names for variables

marina-tilniak (changes requested):

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const rev = {};
```
marina-tilniak, line 28:
```suggestion
  const reversedRobot = {};
```

marina-tilniak (approved):

PR #1447, opened 2022-09-15
Solution

tatyana22:


xiimiik (changes requested):
Try to simplify your, using of nested loops not necessary. 
Check whether there is such a value in the object, if there is - exchange, if not, then return null.
Use .hasOwnProperty()

tatyana22:
> Try to simplify your, using of nested loops not necessary. Check whether there is such a value in the object, if there is - exchange, if not, then return null. Use .hasOwnProperty()

Ok! I want to ask. May be I loose something, in theory I have not seen the theory about hasOwnProperty...

xiimiik:
> Ok! I want to ask. May be I loose something, in theory I have not seen the theory about hasOwnProperty...

https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
For example:
```
if (inversedRobot.hasOwnProperty(robot[key])) {
      return null;
    }
```

tatyana22:
> > Ok! I want to ask. May be I loose something, in theory I have not seen the theory about hasOwnProperty...
> 
> https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty For example:
> 
> ```
> if (inversedRobot.hasOwnProperty(robot[key])) {
>       return null;
>     }
> ```
Yes, I find this. But in theory in our course. https://mate.academy/learn/javascript-advanced/object-advanced#/theory

tatyana22:
> > Ok! I want to ask. May be I loose something, in theory I have not seen the theory about hasOwnProperty...
> 
> https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty For example:
> 
> ```
> if (inversedRobot.hasOwnProperty(robot[key])) {
>       return null;
>     }
> ```

I made with the help of hasOwnProperty. Please, see. Thanks!

xiimiik (changes requested):
create a variable for `robot[key]` you use it twice in code
and
i think it will better to use `key`, than `key1`

tatyana22:
> e a variable for `robot[key]`

OK! Done

xiimiik (approved):
Great!!! GJ 🔥 💯

Andr1yK (approved):

PR #1448, opened 2022-09-15
Solution

GizemEce:


fatihsamur (approved):
Good job 👍 Approved

ihor-jpeg (changes requested):
testing spreadsheet sync

ihor-jpeg (approved):
testing spreadsheet sync

ihor-jpeg (commented):
imagine robot has millions of properties. why would you first do all these operations and only then return null if values repeat. maybe you can do it during the iteration? it's more efficient

ihor-jpeg (changes requested):
.

ihor-jpeg (approved):

PR #1449, opened 2022-09-15
Solution

BZOzyurt:


Quelaan1 (approved):
Good job.

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const correctedRobots = {};
+
+  for (const key in robot) {
+    correctedRobots[robot[key]] = key;
+  }
+
+  if (Object.keys(correctedRobots).length !== Object.keys(robot).length) {
```
ihor-jpeg, line 33: too hard to read. make variables. but you actually do not need to use this. return null from the for loop

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #1452, opened 2022-09-15
solution

WebsterManyanga:


Bibicutu (approved):
You could avoid nested `for`, but overall, good job!

src/inverseRobot.js
```diff
@@ -25,6 +25,24 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+
+  for (let i = 0; i < keys.length - 1; i++) {
+    for (let j = i + 1; j < keys.length; j++) {
```
ihor-jpeg, line 32: nope. too expensive of a calculation. you really need one for loop here and one object

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #1453, opened 2022-09-16
Solution for inverse Robot

sapnachoudhary06:


Bibicutu (approved):
Nice solution! Good job!

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+  const uniqueKeys = [];
+
+  for (const key in robot) {
+    if (uniqueKeys.includes(robot[key])) {
+      return null;
```
ihor-jpeg, line 33: let's not return from the for loop
you can use the `continue` directive

ihor-jpeg: oh, sorry, the function needs to stop executing. that's correct in this case

ihor-jpeg (approved):

PR #1454, opened 2022-09-16
solution

bokorzita:


Bibicutu (approved):
Good job!

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+
+  if (uniqueValues.size < values.length) {
+    return null;
+  }
+
+  const swapped = Object.entries(robot).map(
+    ([key, value]) => ({ [value]: key })
+  );
+
+  return Object.assign({}, ...swapped);
```
ihor-jpeg, line 43:
```suggestion
  return Object.entries(robot).map(
    ([key, value]) => ({ [value]: key })
  );
```

ihor-jpeg (commented):
Your solution works well. But imagine we have a very big object, with many keys. Your solution would need to loop over all the keys and only then return null is the values repeat.
i'm pretty sure you can implement it so that it will return null from the loop

ihor-jpeg (approved):

PR #1458, opened 2022-09-17
Solution

omerfarukbalik:


Bibicutu (approved):
Simple and nice solution! Good work!

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  for (const key in robot) {
+    repairedRobot[robot[key]] = key;
+
+    if (repeatedValueFinder.includes(robot[key])) {
+      return null;
+    }
+
+    repeatedValueFinder.push(robot[key]);
+  }
```
ihor-jpeg, line 38: why not simply check if that key already exists in the object? you don't need the array here

ihor-jpeg (approved):

PR #1467, opened 2022-09-19
Solution

MeshackDev:


kamal-tufekcic (approved):
Bit overcomplicated but if it works it works!

ihor-jpeg (changes requested):
use one for loop, not two

ihor-jpeg (approved):

PR #1471, opened 2022-09-20
Solution

Vasyliy22:


SanyaBratashchuk (changes requested):
simplify the solution `fromEntries` and `keys` is redundant here use `for..in` loop

PR #1474, opened 2022-09-21
firt try

metinbicaksiz:


src/inverseRobot.js
```diff
@@ -23,8 +23,24 @@
...
+  for (const key of arrRobot) {
+    mySet.add(key);
+  }
+
+  if (arrRobot.length === mySet.size) {
+    swappedObject = Object.fromEntries(
+      Object.entries(robot).map(v => v.reverse()));
+  } else {
+    return null;
+  }
```
amitpatiljc, line 41: have this check first, and then change key and value exchange step
and just need one condition, 
```suggestion
  if (uniqueValues !== arrRobot.length) {
    return null;
  }
```

src/inverseRobot.js
```diff
@@ -23,8 +23,24 @@
...
+
 function inverseRobot(robot) {
-  // write code here
+  let swappedObject = [];
+  const mySet = new Set();
+  const arrRobot = Object.values(robot);
+
+  for (const key of arrRobot) {
+    mySet.add(key);
+  }
```
amitpatiljc, line 34: this step is redundant

src/inverseRobot.js
```diff
@@ -23,8 +23,24 @@
  *
  * @return {object}
  */
+
 function inverseRobot(robot) {
-  // write code here
+  let swappedObject = [];
+  const mySet = new Set();
+  const arrRobot = Object.values(robot);
```
amitpatiljc, line 30: you can get values and create a set of unique values
```suggestion
  const values = Object.values(robot);
  const uniqueValues = new Set(values).size;
```

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job!

PR #1498, opened 2022-10-11
first commit

haduigon:


src/inverseRobot.js
```diff
@@ -24,7 +24,22 @@
...
+    }
+  }
+
+  for (const key in robot) {
+    robot[robot[key]] = key;
+    delete robot[key];
+  }
+
+  return robot;
 }
```
marina-tilniak, line 43: you need only one for in loop here(and hasOwnProperty for checking)

marina-tilniak (changes requested):

src/inverseRobot.js
haduigon: @marina-tilniak , I have been thinking about it whole week and still have no idea how to solve with one one for in loop and hasOwnProperty, need some hint )

haduigon: it is done )

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  // const arr = Object.values(robot);
+
```
TarasHoliuk, line 28: comment and empty line are redundant

TarasHoliuk (changes requested):
GJ! but I've added one comment - fix it please

src/inverseRobot.js
haduigon: @TarasHoliuk , fixed, sorry )

TarasHoliuk: you have nothing to apologize for 🙂

TarasHoliuk (approved):

PR #1504, opened 2022-10-15
JS_INVERSE_ROBOT TASK SOLUTION

divertitoS:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+
+  if (new Set(values).size !== values.length) {
+    return null;
+  }
```
Bohdan-Maksymiuk, line 31: good, but instead of this check we could check if fixedRobot already has such key during iteration
but need to calculate time and space complexity

Bohdan-Maksymiuk (approved):

PR #1505, opened 2022-10-15
Task was done

Sergey-Mironenko:


src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robot2 = {};
+
+  for (let i = 0; i < Object.keys(robot).length; i++) {
```
loralevitska, line 29: DRY 
`Object.keys(robot)` - move to variable
`Object.values(robot)` - move to variable

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robot2 = {};
```
loralevitska, line 27:
```suggestion
  const inversedRobot = {};
```

loralevitska (changes requested):
great job) 
It remains to improve your code style

src/inverseRobot.js
```diff
@@ -24,7 +24,25 @@
...
+  const robotKeys = Object.keys(robot);
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotKeys.length; i++) {
+    for (let j = i + 1; j < robotKeys.length; j++) {
+      if (robotValues[i] === robotValues[j]) {
+        return null;
+      }
+    }
+  }
```
tetiana-ivanchenko, line 37: This nested loop is redundant, it is possible to solve this task using only one of your forEach loops.
Let'ssimplify your code!)
if you want to avoid duplicates you can just not include the already existing property in the new object. Inside your forEach callback: 
if you have this property in your new object, return null, 
and only if not, add such property to the new object.

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):
GJ!

PR #1506, opened 2022-10-17
solution

MaxFedak:


Kry-Gin (approved):
Well done!

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const robotInversed = {};
```
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+  const valuesOldRobot = Object.values(robot);
```
Olexandr6, line 29:
```suggestion
  const oldRobotValues = Object.values(robot);
```

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+  const keysOldRobot = Object.keys(robot);
```
Olexandr6, line 30: Here is the same)

src/inverseRobot.js
```diff
@@ -25,6 +25,19 @@
...
+
+  for (let i = 0; i < valuesOldRobot.length; i++) {
+    if (Object.keys(robotInversed).includes(valuesOldRobot[i])) {
+      return null;
```
Olexandr6, line 34: Let's find another way to check if a property is an object

Olexandr6 (changes requested):
Your solution works, but it can be improved

Olexandr6 (approved):
Looks much better🔥

PR #1521, opened 2022-10-18
Solution

yaro-bey:


Anna-Shapovalova (approved):
GJ) Looks like everything is OK) U can make it a little bit easier (without array), but I do like your solution) :relaxed:

mykola-koval (approved):

PR #1524, opened 2022-10-19
solution

IOneStarI:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (fixedRobot[robot[key]]) {
```
mitattt, line 30: you use robot[key] several times so you should add this in variable

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+      return null;
+    }
+
+    fixedRobot[robot[key]] = key;
```
mitattt, line 34: same here

src/inverseRobot.js
mitattt, line 30: its better to use `hasOwnProperty` rather than this >` fixedRobot[robot[key]]` CHECKLIST.2

mitattt (changes requested):
GJ just a few mistakes

src/inverseRobot.js
IOneStarI: 2 times i think all right)

Olexandr6 (approved):

PR #1526, opened 2022-10-19
inverseRobot task is done

AlexTsimba:


spiritkyiv (commented):
I think, use Object.entries and then iterate it to reverse key and value in obj is unjustified

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const arr = Object.entries(robot);
+  const result = {};
+
+  for (let i = 0; i < arr.length; i++) {
+    result[arr[i][1]] = arr[i][0];
+  }
```
mykola-koval, line 32: use `for in` to iterate over objects

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const arr = Object.entries(robot);
+  const result = {};
```
mykola-koval, line 28: bad var names

mykola-koval (changes requested):

src/inverseRobot.js
```diff
@@ -19,12 +19,19 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const property in robot) {
+    if (!inversedRobot.hasOwnProperty([robot[property]])) {
+      inversedRobot[robot[property]] = property;
+    } else {
+      return null;
+    }
```
illiarerun, line 31: invert your if statement, that way you won't need `else`, try to avoid eles
```suggestion
    if (inversedRobot.hasOwnProperty([robot[property]])) {
      return null;
    } 
    
    inversedRobot[robot[property]] = property;
```

illiarerun (approved):

PR #1528, opened 2022-10-19
solution

DanilSerzh:


src/inverseRobot.js
```diff
@@ -24,7 +24,27 @@
...
+  }
+
+  for (let i = 0; i < Object.keys(newRobot).length; i++) {
+    newCounter++;
+  }
+
+  if (oldCounter === newCounter) {
+    return newRobot;
+  } else {
+    return null;
```
taraselnik, line 46: they asked me not to write 'else' i see it can be change to avoid using else)

mykola-koval (approved):

PR #1531, opened 2022-10-19
solution

Yurii-Pozho:


src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let result = {};
+
+  for (const key in robot) {
+    if (result[robot[key]]) {
+      result = null;
+      break;
```
mitattt, line 32: why do you need this `break`?

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+    } else {
```
mitattt, line 33: try not to use if | else construction

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  let result = {};
```
mitattt, line 27: 3.[NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)

src/inverseRobot.js
```diff
@@ -24,7 +24,18 @@
...
+
+  for (const key in robot) {
+    if (result[robot[key]]) {
```
mitattt, line 30: you use robot[key] several times so you should add this in variable

mitattt (changes requested):
It works but you can do it more readable

src/inverseRobot.js
Yurii-Pozho: when I find a repeated object, I don't need to look for it anymore

PR #1534, opened 2022-10-19
Solution inversus

SerhiiVovchenko:


src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newObject = {};
```
Anastasiia-Svintsova, line 27: it is better to name variables with meaningful names
```suggestion
  const newRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  for (const keys in robot) {
+    const newKey = keys;
```
Anastasiia-Svintsova, line 30: Are you sure you need a new const for keys? Why do you can't use keys?

Anastasiia-Svintsova (changes requested):
Good! Just fix some comments

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const Key in robot) {
```
Bohdan-Maksymiuk, line 29:
```suggestion
  for (const key in robot) {
```
LowerCase for variable name

Bohdan-Maksymiuk (approved):

PR #1535, opened 2022-10-19
Develop

AndreyChernysh:


src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotRevert = {};
+
+  for (const key in robot) {
+    const path = robot[key];
```
Olexandr6, line 30: What is `path`)?
```suggestion
    const robotKeyValue = robot[key];
```

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const robotRevert = {};
```
Olexandr6, line 27:
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+
+  for (const key in robot) {
+    const path = robot[key];
+
+    if (robotRevert[path]) {
```
Olexandr6, line 32: What if `undefined` or `null` are the values of this object?

Find another way how to check if a property exists in an object

Olexandr6 (changes requested):

Olexandr6 (approved):
Gj!

PR #1536, opened 2022-10-19
Solution

MikhailKucherina:


src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  const values = Object.values(robot);
+
+  const sortedValues = values.sort();
+
+  for (let i = 0; i < sortedValues.length - 1; i++) {
```
shtht24, line 33: try even better example from cl, it's more simple than making another one cycle

shtht24 (changes requested):

Olexandr6 (approved):
Well done🔥

PR #1538, opened 2022-10-19
Solution

sofiia-dovhan:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
```
mitattt, line 30: you use robot[key] several times so you shoud add this in variable

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
...
+      return null;
+    }
+
+    newRobot[robot[key]] = key;
```
mitattt, line 34: same here

mitattt (changes requested):

src/inverseRobot.js
sofiia-dovhan: I don't think it is necessary here

src/inverseRobot.js
sofiia-dovhan: same here

Olexandr6 (approved):
Good job!

PR #1539, opened 2022-10-19
Solution

nazarko94:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    if (!inverse[robot[key]]) {
+      inverse[robot[key]] = key;
+    } else {
```
Oleksandr-Stryzhavchuk, line 32: I think it would be better to use 'hasOwnProperty', so as not to use 'else' later, but it is only my opinion.

Oleksandr-Stryzhavchuk (approved):
good job

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    if (!inverse[robot[key]]) {
```
mykola-koval, line 30: checklist 2

mykola-koval (changes requested):

mykola-koval (approved):

PR #1540, opened 2022-10-19
Solution was Created

Stepico:


mykola-koval (changes requested):
veeeery difficult solution :)
let's try to make it simpler, make some student reviews to get an inspiration

mykola-koval (approved):

PR #1543, opened 2022-10-19
Solution

Oleksandr-Stryzhavchuk:


nazarko94 (approved):
@Oleksandr-Stryzhavchuk gj, I think it's correct way I did the same)

PR #1546, opened 2022-10-19
 js_inverse_robot

MXMUK:


nazarko94 (approved):
Wouldn't be better to use loop for in to work with objects?

MXMUK:
yeah you're right, but i realized it after I had already done it. Now I don't use this cycle for objects 😅

src/inverseRobot.js
```diff
@@ -25,6 +25,17 @@
  */
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const key of Object.keys(robot)) {
```
mykola-koval, line 30: use `for in`

mykola-koval (changes requested):

mykola-koval (approved):

PR #1547, opened 2022-10-19
Solution

andrii-bassarab:


referee1988 (approved):
Great job!

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
-}
+  const reverse = {};
```
Olexandr6, line 27:
```suggestion
  const reversedRobot = {};
```

Olexandr6: Pay attention to variable naming

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+  const entries = Object.entries(robot);
```
```suggestion
  const robotEntries = Object.entries(robot);
```

src/inverseRobot.js
```diff
@@ -24,7 +24,20 @@
...
+
+  for (const object of entries) {
+    const value = object[1];
+    const key = object[0];
```
Olexandr6, line 32: You can simplify this by using the destructive assignment:
```suggestion
  for (const [key, value] of entries) {
```

Olexandr6 (changes requested):
Pay attention to variable naming!

src/inverseRobot.js
andrii-bassarab: thanks for the advice, it's really easier and better

denys-danyliuk (approved):

PR #1548, opened 2022-10-19
Solution

referee1988:


src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const goodRobot = {};
+  let isWrongRobot = false;
+
+  for (const key in robot) {
+    if (goodRobot.hasOwnProperty(robot[key])) {
```
andrii-bassarab, line 31: you use "robot[key]" several times so you should add this in variable

andrii-bassarab (commented):
Good job, but fix this point

src/inverseRobot.js
referee1988: Variables are needed for the result of an expression that is repeated or for the result of a function that needs to be used multiple times.

robot[key] is a variable (link, pointer) that contains the address of the key property of the robot object. 

Why assign one variable the value of another variable?

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+      isWrongRobot = true;
+      break;
```
mykola-koval, line 33: instead of using unnecessary variables you can return null here
```suggestion
      return null
```

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const goodRobot = {};
+  let isWrongRobot = false;
```
mykola-koval, line 28:
```suggestion
```

src/inverseRobot.js
```diff
@@ -24,7 +24,23 @@
...
+      isWrongRobot = true;
+      break;
+    }
+
+    goodRobot[robot[key]] = key;
+  }
+
+  if (isWrongRobot) {
+    return null;
+  }
```
mykola-koval, line 41:
```suggestion
  }
```

mykola-koval (changes requested):

mykola-koval (approved):

PR #1561, opened 2022-10-26
Solution

Oleksandr-Filo:


Fynkzia (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+  const robotEntries = Object.entries(robot);
+  const corectRobot = {};
+
+  for (const data of robotEntries) {
+    if (!corectRobot[data[1]]) {
+      corectRobot[data[1]] = data[0];
+    } else {
+      return null;
+    }
+  }
```
valerii-derkach, line 19: checklist p2, 4

valerii-derkach (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotEntries = Object.entries(robot);
```
solaryasha, line 10: You have a working solution, but let me ask you: do you really need to store in the memory one array of robot entries?

Maybe could check the key in the newly created object?

solaryasha (changes requested):

src/inverseRobot.js
Oleksandr-Filo: I agree, no need to create a new array, fixed

solaryasha (approved):
👍

PR #1573, opened 2022-10-30
test Ok

diolna:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+  const arrValues = Object.values(robot);
+  const arrKeys = Object.keys(robot);
+
+  for (const key in robot) {
+    if (arrValues.filter(n => n === robot[key]).length > 1
+    || arrKeys.filter(n => n === key).length > 1) {
+      return null;
+    }
+    fixRobot[robot[key]] = key;
+  }
```
nazarmatsevych, line 21:
```suggestion
    }
    
    fixRobot[robot[key]] = key;
  }
```

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const fixRobot = {};
+
+  const arrValues = Object.values(robot);
+  const arrKeys = Object.keys(robot);
+
+  for (const key in robot) {
+    if (arrValues.filter(n => n === robot[key]).length > 1
+    || arrKeys.filter(n => n === key).length > 1) {
```
nazarmatsevych, line 17: You only need condition for values.
You also can use [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) method instead of Object.values and filter

nazarmatsevych (changes requested):
Good work
Check comments pls

src/inverseRobot.js
diolna: Fixed. Did you mean it?
And still another option. Check after reverse:

  if (Object.keys(fixRobot).length
    !== Object.keys(robot).length) {
    return null;
  }

PR #1576, opened 2022-10-31
inverseRobot

Mykola-Busarhin:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const nameSake = {};
+
+  for (const key in robot) {
+    if (!nameSake[robot[key]]) {
+      nameSake[robot[key]] = key;
+    } else {
+      return null;
+    }
```
DarkMistyRoom, line 17:
```suggestion
    if (nameSake[robot[key]]) {
      return null;
    }
    
    nameSake[robot[key]] = key;
```

DarkMistyRoom: Try to avoid unnecessary nesting

DarkMistyRoom (approved):
Good, but little suggestion

PR #1577, opened 2022-10-31
Solution

GabrielMaksimovich:


shyptia (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newInverseObj = {};
```
solaryasha, line 10: Avoid using data structure names in variable names, it's really confusing
```suggestion
  const inversedRobot = {};
```

solaryasha (approved):

PR #1578, opened 2022-10-31
Solution inverseRobot

vadiimvooo:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInverse = {};
+  let name;
+
+  for (const key in robot) {
+    name = robot[key];
```
llllewvvaa, line 14: I think it would be better if you implement the variable here, and use `const` instead of `let`
```suggestion
    const name = robot[key];
```

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInverse = {};
+  let name;
```
llllewvvaa, line 11: delete
```suggestion
```

llllewvvaa (changes requested):

llllewvvaa (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInverse = {};
```
```suggestion
  const inversedRobot = {};
```

solaryasha (approved):
👍

PR #1589, opened 2022-11-07
Solution

RomaSheva1987:


src/inverseRobot.js
```diff
@@ -7,7 +7,31 @@
...
+    for (let j = i + 1; j < arrValues.length; j++) {
+      if (arrValues[i] === arrValues[j]) {
+        return null;
+      }
+    }
+  }
+
+  for (let i = 0; i < arrKeys.length; i++) {
+    inversedRobot[arrValues[i]] = arrKeys[i];
+  }
```
tetiana-ivanchenko, line 32: do you really need 5 loops, with two of them nested? Actually this task can be solved with just one loop

tetiana-ivanchenko (changes requested):

RomaSheva1987:
this is the shortest solution that i can find out

nazarmatsevych (approved):
Good job!

Instead of Object.keys and Object.values, you can use for ... in cycle and  iterate over the robot object and write keys as values ​​and values ​​as keys in the inversedRobot object

PR #1592, opened 2022-11-07
inverseRobot v1

JeniaGaidukov:


src/inverseRobot.js
```diff
@@ -4,10 +4,22 @@
...
+
+  for (const key of toArray) {
+    result[key[1]] = key[0];
+  }
+
+  if (Object.keys(result).length < Object.values(robot).length) {
+    return null;
+  } else {
+    return result;
+  }
```
qoa11a, line 22: You can iterate over object keys without converting it into array. Also, you can return null as soon as you know that there's already such key.
Your solution works but it might be better. Try to figure out another solution according to my comment.

qoa11a (changes requested):

src/inverseRobot.js
JeniaGaidukov: Випадково зробив commit з кривою назвою, вибачаюсь..

tetiana-ivanchenko (approved):

PR #1593, opened 2022-11-09
inverseRobot

weikmaks:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  const revers = arr.map(function(i) {
+    return i.reverse();
+  });
+
+  const objRevers = Object.fromEntries(revers);
+
+  if (Object.keys(objRevers).length !== Object.keys(robot).length) {
+    return null;
+  }
```
DarkMistyRoom, line 20: will be better to do it in beginning to avoid unnecessary operations

DarkMistyRoom (approved):
It works, but it isn't the best for performance.
something like this will work faster
image

PR #1595, opened 2022-11-10
fSolution

yurii-filiushkin:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+    if (values.includes(values[index], index + 1)) {
+      newRobot = null;
+
+      return newRobot;
+    }
+  }
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
```
tetiana-ivanchenko, line 24: can you combine two loops into one?
hint: to check If any of the object values are repeated, you can check if robot[key] is already exist in newRobot in the secons loop

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
yurii-filiushkin: Oh, really much better solution. Thank

nazarmatsevych (approved):
Good work!

PR #1599, opened 2022-11-13
Initial commit

artem-hirzhev:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    result[robot[key]] = key;
```
ivanchuk-yurii, line 13: it's better to check here if `result[robot[key]]` already exists and return `null`

ivanchuk-yurii (approved):

PR #1602, opened 2022-11-13
Solution

panianchuk:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invRobot = {};
+
+  for (const key of Object.keys(robot)) {
+    invRobot[robot[key]] = key;
```
ivanchuk-yurii, line 13: it's better to check here if `invRobot` has `robot[key]` and return `null`

ivanchuk-yurii (approved):

Bohdan-Maksymiuk (approved):

PR #1604, opened 2022-11-14
function inverseRobot was implemented

Chupiak:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const values = Object.values(robot);
+  const newRobot = {};
+
+  for (let i = 0; i < keys.length; i++) {
+    newRobot[values[i]] = keys[i];
+  }
+
+  if (Object.keys(newRobot).length !== keys.length) {
+    return null;
+  }
```
Bohdan-Maksymiuk, line 20: better to check inside loop if key exists instead of check after full iteration

Bohdan-Maksymiuk (changes requested):
Good job, lets improve performance of your code

Bohdan-Maksymiuk (changes requested):
I didn't see any check inside the loop - check if there is such key in object and return null from function there

src/inverseRobot.js
Chupiak: Please check

DarkMistyRoom (approved):
Good!

PR #1605, opened 2022-11-14
Solution

OksanaBaloh:


src/inverseRobot.js
```diff
@@ -8,6 +8,21 @@
...
+  if (keys.length !== uniqueKeys.size) {
+    return null;
+  }
+
+  const values = Object.keys(robot);
+  const invertedRobot = {};
+
+  for (let i = 0; i < keys.length; i++) {
+    invertedRobot[keys[i]] = values[i];
+  }
```
Bohdan-Maksymiuk, line 23: why do you need separate array for object keys? (and why did you name the variable `values`?)
you van iterate by keys using `for in` loop

Bohdan-Maksymiuk (changes requested):

src/inverseRobot.js
OksanaBaloh: yes, i agree. It was strange solution) I also changed code, where null was returned

Bohdan-Maksymiuk (approved):

PR #1606, opened 2022-11-14
Solution

OleksandrOse:


src/inverseRobot.js
```diff
@@ -8,6 +8,14 @@
...
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const val in robot) {
+    newRobot[robot[val]] = val;
+  }
+
+  return Object.keys(newRobot).length < Object.keys(robot).length
+    ? null : newRobot;
```
Bohdan-Maksymiuk, line 18: good, but better to check if key existed in loop
Then you will not iterate through whole array, and return from function when you need it

Bohdan-Maksymiuk (changes requested):

src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
...
   // write code here
+  let newRobot = {};
+
+  for (const detail in robot) {
+    if (newRobot.hasOwnProperty(robot[detail])) {
+      newRobot = null;
+      break;
+    } else {
+      newRobot[robot[detail]] = detail;
+    }
```
Bohdan-Maksymiuk, line 19:
```suggestion
    if (newRobot.hasOwnProperty(robot[detail])) {
      return null;
    }
    
    newRobot[robot[detail]] = detail;
```
if you will return from function - loop will break too

Bohdan-Maksymiuk (approved):

PR #1607, opened 2022-11-14
Solution

Dmytro-Litinskyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotReverse = {};
+  const rKeys = Object.keys(robot);
+  const rValues = Object.values(robot);
```
tetiana-ivanchenko, line 12: redundant if you will use "for in" in robot

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  for (let i = 0; i < rKeys.length; i++) {
```
tetiana-ivanchenko, line 14: try to use "for in" in robot

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1609, opened 2022-11-15
commit message

jaxxxon78:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const revRobot = {};
+
+  for (const key in robot) {
+    if (!revRobot[robot[key]]) {
+      revRobot[robot[key]] = key;
+    } else {
+      return null;
+    }
```
Bohdan-Maksymiuk, line 17: try to avoid `if else`
```suggestion
    if (revRobot[robot[key]]) {
      return null;
    }
    
    revRobot[robot[key]] = key;
```

Bohdan-Maksymiuk: and better to use `hasOwnProperty()`

Bohdan-Maksymiuk (changes requested):

Viktor-Kost (approved):
Nice!

PR #1610, opened 2022-11-15
solution

Fynkzia:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    if (inverse.hasOwnProperty([robot[key]])) {
+      return null;
+    } else {
+      inverse[robot[key]] = key;
+    }
```
Ihor-Karpyn, line 17: p3 - https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

It's not this task's checklist, but still, this advice is important

Ihor-Karpyn (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
```
solaryasha, line 10: If the variable name starts with the verb, it should be the function.
You got an object
```suggestion
  const inversedRobot = {};
```

solaryasha (approved):

PR #1612, opened 2022-11-16
Solution done!

IoghiStudio:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotParts = Object.entries(robot);
+  let modified = {};
+  const cache = {};
+
+  for (const part of robotParts) {
+    if (!cache[part[1]]) {
```
mgrinko, line 15: you could check `key in modified` and do not create cache at all

mgrinko (approved):

PR #1613, opened 2022-11-17
Solution

llllewvvaa:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverse = {};
+
+  for (const key in robot) {
+    const property = robot[key];
 
+    if (inverse[property]) {
```
mr-valentos, line 15: [CODE STYLE] - use correct check for property presence in object. Some properties can be present, but still contain falsy value.
GOOD EXAMPLE:

if (key in robot) {
EVEN BETTER EXAMPLE:

BAD EXAMPLE:

if (robot[key] === undefined) {

if (robot[key]) {

mr-valentos (changes requested):
checklist p.2

src/inverseRobot.js
llllewvvaa: **Fixed**

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverse = {};
```
solaryasha, line 10: Sound like a function, let's change the naming for 
```suggestion
  const inversedRobot = {};
```

solaryasha (approved):

PR #1615, opened 2022-11-19
Solution inverseRobot

Sasha-Green:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]] !== undefined) {
```
Anastasiia-Svintsova, line 13:
```suggestion
    if (newRobot[robot[key]]) {
```
or
```suggestion
    if (robot[key] in newRobot) {
```

Anastasiia-Svintsova (changes requested):
Good! Fix a little comment

TarasHoliuk (approved):

PR #1617, opened 2022-11-19
Solution

fallowlone:


mr-valentos (approved):
GJ

official-artem:
> GJ

Thanks for your checking.

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const values = Object.values(robot);
+  const repeatCheak = new Set(values).size === values.length;
```
solaryasha, line 12: You have a working solution, but let me ask you: do you really need to store in the memory one array and the set for getting info about unique values?
Or could you just check the keys in your new object?

solaryasha (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
solaryasha, line 10: the `new` keyword in the variable name don't any meaning
```suggestion
  const inversedRobot = {};
```

solaryasha (approved):

PR #1618, opened 2022-11-19
inverse robot

mr-valentos:


src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
...
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  let newRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
```
official-artem, line 12: You can create a variable with entries separately and use it in your loop;

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
...
+    if (newRobot.hasOwnProperty(value)) {
+      newRobot = null;
```
official-artem, line 14: You can return null right away;

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
...
+      break;
```
official-artem, line 15:
```suggestion
```

official-artem (approved):

src/inverseRobot.js
official-artem: > 

If you return null right away. You should delete 'break'.

solaryasha (changes requested):
Let's fix Artem's comments

src/inverseRobot.js
mr-valentos: let's leave it up to the mentors

solaryasha (approved):
💯

PR #1619, opened 2022-11-19
develop inverse robot

constantweb:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  let inverse = {};
+
+  for (const key in robot) {
+    if (inverse[robot[key]]) {
+      inverse = null;
+      break;
```
TarasHoliuk, line 15:
```suggestion
      return null;
```
We can immediately `return` - it's more transparent, and shorter, no need `let` variable

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #1621, opened 2022-11-20
Solution

halynarydasova:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const setOfUniqueValues = new Set();
+  const arrayOfValues = Object.values(robot);
+
+  arrayOfValues.forEach((elem) => {
+    setOfUniqueValues.add(elem);
+  });
```
jstmpelowycz, line 16: set make them unique itself, no need to indicate it in its name

```suggestion
  const values = Object.values(robot);
  const uniqueValues = new Set(values);
```

jstmpelowycz (changes requested):
almost done!

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const UniqueValues = new Set();
+  const Values = Object.values(robot);
```
jstmpelowycz, line 12: it should not be capitalized (capitalized variables refer to classes, here's just objects
```suggestion
  const uniqueValues = new Set();
  const values = Object.values(robot);
```

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+
+  Values.forEach((elem) => {
+    UniqueValues.add(elem);
+  });
```
jstmpelowycz, line 16: no need to overcomplicate your code, it's possible to pass array to `Set()` to make only unique values stay in there

jstmpelowycz (changes requested):

jstmpelowycz (approved):
great!

PR #1624, opened 2022-11-20
Solution

vlad-boikov:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const result = {};
+
+  const robotKeys = Object.keys(robot);
+
+  for (const key of robotKeys) {
+    if (result[robot[key]] !== undefined) {
+      return null;
+    } else {
+      result[robot[key]] = key;
+    };
```
stanislav-topikha, line 19:
```suggestion
    } 
    
    result[value] = key;
```
you dont need `else` here, use var for value

stanislav-topikha (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+  const result = {};
+
+  const robotKeys = Object.keys(robot);
+
+  for (const key of robotKeys) {
+    const value = robot[key];
+
+    if (result[value] !== undefined) {
+      return null;
+    }
```
SanyaBratashchuk, line 19:
```suggestion
    if (value in result) {
      return null;
    }
```

SanyaBratashchuk (approved):

PR #1627, opened 2022-11-21
solution

ysrckr:


nurdanokcu (approved):
Clear implementation.

src/inverseRobot.js
```diff
@@ -7,7 +7,27 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+
+  if (hasDuplicates(keys)) {
+    return null;
+  }
```
mgrinko, line 16: this situation is not possible!
Keys are always different otherwise the last value just replaces the previous one

mgrinko (approved):

PR #1628, opened 2022-11-21
solution

NadiiaKotliar:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  for (const kay in robot) {
```
vlad-demchuk, line 10:
```suggestion
  for (const key in robot) {
```

vlad-demchuk (changes requested):

tetiana-ivanchenko (changes requested):
image
you should return the new obj, insead of modification incoming obj

AlisaVlasova (approved):

PR #1630, opened 2022-11-22
Solution

UmizDemud:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const obj = Object.entries(robot);
```
mgrinko, line 11: `obj` is a really bad name for an array containing key, value pairs

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (let i = 0; i < obj.length; i++) {
```
mgrinko, line 13: Why don't you just iterate the object keys here?

mgrinko (changes requested):

src/inverseRobot.js
UmizDemud: Yes, Object.keys makes more sense, correcting. I didn't use a forEach because a return in the middle way would be beneficial

src/inverseRobot.js
UmizDemud: Yes, I'm used to naming so that my friends understand :D Correcting quickly

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const robotKeys = Object.keys(robot);
+
+  for (let i = 0; i < robotKeys.length; i++) {
```
mgrinko, line 13:
```suggestion
  for (const key of Object.keys(robot)) {
```

mgrinko (approved):

src/inverseRobot.js
UmizDemud: 👍

PR #1633, opened 2022-11-22
Inverse robot

OlexandraKorol:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
TetKopot, line 10: you could use here proper name of variables

TetKopot (commented):
good job, but write another name for the object

src/inverseRobot.js
OlexandraKorol: Done)

solaryasha: Not done, check the p3 of the checklist

solaryasha (changes requested):

ivan-kadykalo (approved):

PR #1639, opened 2022-11-22
solution

lefteris-L:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+
+  for (const key of keys) {
+    if (keys.lastIndexOf(key) !== keys.indexOf(key)) {
+      return null;
+    }
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    inverse[keys[i]] = values[i];
+  }
```
Lun4trick, line 22:
```suggestion
  for (let i = 0; i < keys.length; i++) {
   if (keys.indexOf(keys[i]) !== keys.lastIndexOf(keys[i])) {
      return null;
    }
  
    inverse[keys[i]] = values[i];
  }
```

i think u dont need 2 separate loops here.

Lun4trick (approved):

meforoma (approved):

PR #1640, opened 2022-11-23
solution

kreisberg:


VladimirEdelmann (approved):
You don't need to add package.json file here. When you do commiting add only src/

solaryasha (approved):
👍

kreisberg:
> You don't need to add package.json file here. When you do commiting add only src/

get it)

PR #1641, opened 2022-11-23
solution

Lun4trick:


lefteris-L (approved):
Excellent 👍

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.keys(robot);
+  const keys = Object.values(robot);
+  let newBot = {};
+
+  for (let i = 0; i < values.length; i++) {
```
meforoma, line 14: it would be much more concise if you use the for in loop:
`for (const key in robot) {` which iterates through object keys
in such a way swapping is as easy as: `newRobot[robot[key]] = key`

meforoma (changes requested):
let's make it much simpler - see the comment

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
-  // write code here
+  const values = Object.keys(robot);
+  const keys = Object.values(robot);
+  let newBot = {};
+
+  for (const key in keys) {
+    if (keys.indexOf(keys[key]) !== keys.lastIndexOf(keys[key])) {
+      newBot = null;
+
+      return newBot;
```
mgrinko, line 18: you don't need this assignment

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.keys(robot);
+  const keys = Object.values(robot);
+  let newBot = {};
+
+  for (const key in keys) {
```
mgrinko, line 14: You iterate an array with `for .. in` which is wrong

mgrinko (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.keys(robot);
+  const keys = Object.values(robot);
+  const newBot = {};
+
+  for (const key of keys) {
+    if (keys.indexOf(key) !== keys.lastIndexOf(key)) {
```
mgrinko, line 15: Great approach but quite slow

mgrinko (approved):

PR #1644, opened 2022-11-23
Robot inversed

grslckr:


src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    inversedRobot[robot[key]] = key;
+  }
+
+  const objectValues = Object.values(robot);
+
+  for (let i = 0; i < objectValues.length; i++) {
+    const firstIndex = objectValues.indexOf(objectValues[i]);
```
lefteris-L, line 19: `objectValues.indexOf(objectValues[i])` is the same as just `i` in this case.

lefteris-L (approved):
Great solution, especially the first loop 👍

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    inversedRobot[robot[key]] = key;
+  }
+
+  const objectValues = Object.values(robot);
```
mgrinko, line 16: you could just compare the number of keys Object.keys(inversedRobot).length

mgrinko (approved):

src/inverseRobot.js
grslckr: wow you are definitely right, i don't know why i didn't think of that 
thank you

grslckr: you are definitely right i don't know why i didn't think of that 😄

PR #1646, opened 2022-11-23
Finished task

JzSol:


lefteris-L (approved):
Short and simple 💯

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
...
 
 function inverseRobot(robot) {
   // write code here
+  const revese = {};
+
+  for (const keys in robot) {
+    if (revese.hasOwnProperty(robot[keys])) {
+      return null;
+    }
+    revese[robot[keys]] = keys;
```
mgrinko, line 17: A blank line between a block and the next command

```suggestion
    }
    
    revese[robot[keys]] = keys;
```

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
...
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
   // write code here
+  const revese = {};
+
+  for (const keys in robot) {
```
mgrinko, line 13: Why do you call a variable containing a current `key` in plural form?

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
  * @param {object} robot
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
   // write code here
+  const revese = {};
```
mgrinko, line 11: `reverse` is a verb
It is a good name for a function but not for a variable storing an object

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
  * @param {object} robot
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
   // write code here
```
mgrinko, line 10: Do you really need this comment?

mgrinko (changes requested):

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
  * @param {object} robot
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
naziamoff, line 10: your variables should have more descriptive names, so you should avoid naming them like `result` or `object`.
What about `inversedRobot`?

naziamoff (approved):
Good job!
Just small comment about naming

PR #1651, opened 2022-11-23
solution

mmahfuzi:


src/inverseRobot.js
```diff
@@ -8,6 +8,29 @@
...
+  const k = [];
+
+  for (const key in robot) {
+    count++;
+    k.push(key);
+  }
+
+  for (let i = 0; i < count; i++) {
+    for (let j = i + 1; j < count; j++) {
+      if (Object.values(robot)[i] === Object.values(robot)[j]) {
```
Lun4trick, line 21:
```suggestion
```
I think its better to define a variable at the start for the object values, cause in this case its recreate the array every time when the loop executes.

src/inverseRobot.js
```diff
@@ -8,6 +8,29 @@
...
   // write code here
+  let count = 0;
+  const k = [];
+
+  for (const key in robot) {
+    count++;
+    k.push(key);
+  }
+
+  for (let i = 0; i < count; i++) {
```
Lun4trick, line 19:
```suggestion

```
You dont need two loops here, you can do this if you cheking some indexes :)

Lun4trick (changes requested):

src/inverseRobot.js
mmahfuzi: but without two loops how can I compare each pair in the object with each and every other pair?

src/inverseRobot.js
```diff
@@ -8,6 +8,29 @@
 
 function inverseRobot(robot) {
   // write code here
+  let count = 0;
+  const k = [];
+
+  for (const key in robot) {
+    count++;
+    k.push(key);
+  }
```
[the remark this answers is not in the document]
mmahfuzi: to iterate the following nested loops I need the count number, the list k is redundant, it is just to avoid an error. The error was for the key being declared but never used.

Lun4trick (approved):
Good job 👍 :)

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const values = Object.values(robot);
+  const result = {};
+
+  for (const key in robot) {
+    if (values.indexOf(robot[key]) !== values.lastIndexOf(robot[key])) {
+      return null;
+    }
```
meforoma, line 17: you can simplify it by just checking `result[robot[key]]` - then you won't need `values` variable at all
```suggestion
    if (result[robot[key]]) {
      return null;
    }
```

meforoma (changes requested):
see just a single optimisation comment

meforoma (approved):
🎉

PR #1662, opened 2022-11-24
completed

alaaashkar:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const obj = {};
```
mgrinko, line 10: `obj` is not the best name for the copied robot :)

mgrinko (approved):

PR #1663, opened 2022-11-24
finished

jackelashkar:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const obj = {};
```
mgrinko, line 10: `obj` is not the best name for the copied robot

mgrinko (approved):

src/inverseRobot.js
jackelashkar: I will take this into consideration Misha!

PR #1665, opened 2022-11-24
Solution

Mohameddahi:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(newRobot).length) {
+    return null;
+  }
```
meforoma, line 18: you can check the property existence inside a loop
```suggestion
  for (const key in robot) {
    if (newRobot[robot[key]]) {
      return null;
    }

    newRobot[robot[key]] = key;
  }
```

meforoma (changes requested):
check out an optimisation comment

meforoma (approved):

PR #1671, opened 2022-11-24
Solution

oleh-yablunovskyi:


VladimirEdelmann (approved):
Great!

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
```suggestion
  const inversedRobot = {};
```
Avoid using new word in the variable naming, it's a kind of confusing and don't bring any meaning whatsoever

solaryasha (approved):

src/inverseRobot.js
oleh-yablunovskyi: Done!

PR #1673, opened 2022-11-24
Inverse robot

YakovlewDan:


VladimirEdelmann (changes requested):
Don't add any files when you do committing, but those in src/ folder
Try to implement solution only with one for in loop and hasOwnProperty method, see checklist p2

VladimirEdelmann (approved):

Ihor-Karpyn (approved):

PR #1676, opened 2022-11-25
Solution

kattentionn:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotRight = {};
```
VladimirEdelmann, line 10: What means robotRight? Why not robotLeft? =) reversedRobot is clearer

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const key in robot) {
+    if (robot[key] in robotRight) {
```
VladimirEdelmann, line 13: Good! For some reason hasOwnProperty is better. See checklist p2

VladimirEdelmann (approved):

src/inverseRobot.js
kattentionn: I meant a different meaning, but you're right, your name is better)

src/inverseRobot.js
kattentionn: Yes, I saw it in the checklist, but in principle I will leave this option, it fits a good example

VladimirEdelmann (approved):

PR #1677, opened 2022-11-25
DiDemkiv solution

DiDemkiv:


anastasiia-bilska (approved):
Everything seems good to me!

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newInverseRobot = {};
```
```suggestion
  const inversedRobot = {};
```

solaryasha: Avoid using `new` word in the variable naming, it's a kind of confusing

solaryasha (approved):
👍

src/inverseRobot.js
DiDemkiv: fixed

PR #1697, opened 2022-11-28
Solution

dissolveyo:


src/inverseRobot.js
```diff
@@ -4,10 +4,25 @@
...
+  const values = Object.values(robot);
+  const inversedRobot = {};
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = i; j < values.length - i; j++) {
+      if (values[i] === values[j + 1]) {
+        return null;
+      }
+    }
+  }
```
Bohdan-Maksymiuk, line 19: why do you need this nested loops for check?

src/inverseRobot.js
```diff
@@ -4,10 +4,25 @@
...
+
+  for (const key in robot) {
+    inversedRobot[robot[key]] = key;
```
Bohdan-Maksymiuk, line 22: you may check it here if the key already exists in inverseRobot

Bohdan-Maksymiuk (changes requested):

src/inverseRobot.js
```diff
@@ -4,10 +4,22 @@
...
+
+  for (const key in robot) {
+    const inversedRobotKeys = Object.keys(inversedRobot);
+
+    if (inversedRobotKeys.includes(robot[key])) {
+      return null;
+    }
+
+    inversedRobot[robot[key]] = key;
+  }
```
SanyaBratashchuk, line 20:
```suggestion
    if (robot[key] in inversedRobotKeys) {
      return null;
    }

    inversedRobot[robot[key]] = key;
  }
```

SanyaBratashchuk (changes requested):

AlisaVlasova (approved):

PR #1699, opened 2022-11-29
commit one by Edriss Miakhel

Edriss572:


naziamoff (changes requested):
let's use object methods in this task

Edriss572:
> let's use object methods in this task

you mean setter and getter?

PR #1703, opened 2022-12-01
Initial commit

yaroslav-furhala:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const yetanotherRobot = {};
```
Bohdan-Maksymiuk, line 10:
```suggestion
  const yetanotherRobot = {};
```
use camelCase for variables naming and meaningful one

PR #1707, opened 2022-12-02
first attempt

spencerspe:


src/inverseRobot.js
```diff
@@ -5,9 +5,31 @@
...
+  let newRobot = {};
+  const passedValues = [];
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+
+  for (let i = 0; i < keys.length; i++) {
+    newRobot[values[i]] = keys[i];
+  }
+
+  for (let i = 0; i < values.length; i++) {
```
polosanya, line 20: Do you really need one more loop? Can't you do all the checks in the previous one?
Also, what's the purpose of `passedValues` array? You can search for duplicated values if `newRobot`. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

src/inverseRobot.js
```diff
@@ -5,9 +5,31 @@
...
+    }
+    passedValues.push(values[i]);
+  }
+
+  return newRobot;
 }
 
 module.exports = inverseRobot;
+
+// const kolli = { Kolli: 'name', 123: 'chipVer', 3: 'wheels' };
```
polosanya, line 32: Don't forget to delete the comments that you don't need anymore

polosanya (changes requested):
Almost perfect! Please check the comments to improve your code a bit)

polosanya (approved):
Well done! 🔥

PR #1711, opened 2022-12-06
Initial commit

Dynamizx:


jstmpelowycz (changes requested):
generally, you could have approached it in a much easier way
```
  const inversedRobot = {};

  for (const fakeKey in robot) {
    const properKey = robot[fakeKey];

    if (/* if <properKey> exists in <inversedRobot> */) {
      return null;
    }

    // write to <inversedRobot>
  }

  ...
}
```

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const invRobot = {};
+
+  for (const key in robot) {
+    invRobot[robot[key]] = key;
+  }
+
+  if (Object.keys(invRobot).length !== Object.keys(robot).length) {
+    return null;
```
polosanya, line 17: Interesting solution 🔥

polosanya (approved):
Good job!

PR #1716, opened 2022-12-08
Solution

0xNazarii:


andrieieva-yuliia (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+    const newKey = robot[key];
+
+    if (newRobot.hasOwnProperty(newKey)) {
+      return null;
+    }
+
+    newRobot[newKey] = key;
+  }
+
+  return newRobot;
```
alexandra-protyanova, line 22:
```suggestion
  for (const key in robot) {
    if (newRobot.hasOwnProperty(newKey)) {
      return null;
    }
    
    const newKey = robot[key];

    newRobot[newKey] = key;
  }

  return newRobot;
```

alexandra-protyanova: It is pointless to create this variable before, of `return null`)

alexandra-protyanova (approved):
Well done 👍

PR #1720, opened 2022-12-10
Task was done

RomanTsvindin:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+  const robotValues = Object.values(robot);
+  const robotEnries = Object.entries(robot);
+  const unique = Array.from(new Set(robotValues));
+  const reverseEnries = robotEnries.map(a => a.reverse());
+
+  if (unique.length < robotValues.length) {
+    return null;
+  } else {
+    return (Object.fromEntries(reverseEnries.reverse()));
+  }
```
vlad-demchuk, line 19: let's simplify it using one loop🙂
it's too complicated now

vlad-demchuk (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+    return null;
+  } else {
+    for (const key in robot) {
+      const newKey = robot[key];
+
+      invertedRobotEntries.push([newKey, key]);
+    }
+
+    return Object.fromEntries(invertedRobotEntries.reverse());
+  }
```
Anastasiia-Svintsova, line 24: You could avoid nesting. The function stops executing after the return statement anyway
```suggestion
  if (unique.length < robotValues.length) {
    return null;
  }
  
  for (const key in robot) {
    const newKey = robot[key];

    invertedRobotEntries.push([newKey, key]);
  }

  return Object.fromEntries(invertedRobotEntries.reverse());

```

Anastasiia-Svintsova (approved):

PR #1722, opened 2022-12-11
Solution

Vadivel-Babu:


src/inverseRobot.js
```diff
@@ -4,10 +4,30 @@
...
+ */
 
 function inverseRobot(robot) {
   // write code here
+  const values = Object.values(robot);
+
+  const newRobot = {};
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = i + 1; j < values.length; j++) {
```
polosanya, line 16: You can iterate the keys of robot using [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) to avoid nested loops

src/inverseRobot.js
```diff
@@ -4,10 +4,30 @@
...
+      if (values[i] === values[j]) {
+        return null;
+      }
+    }
+  }
+
+  Object.entries(robot).forEach((e) => {
```
polosanya, line 23: Do you really need to iterate all the properties again? Can't you do everything with only one loop?

polosanya (changes requested):
Almost done! Let's try to improve your code a bit

polosanya (approved):
Well done!

PR #1723, opened 2022-12-11
Solution

zvir91:


loralevitska (changes requested):
almost perfect)
try to avoid 2 loops
use [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set?retiredLocale=uk)

You can check `keys.length` and `set.size` and if they are not equal, then immediately return null

zvir91:
> 

thanks

loralevitska (approved):
perfect)

PR #1724, opened 2022-12-11
solution

MrFox32:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const res = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+
+  for (const r in robot) {
```
polosanya, line 12: Same here. What is `r`? Let's use more informative names)

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+    if (res[robot[r]] !== undefined) {
```
polosanya, line 13: image

polosanya (changes requested):
Almost perfect! Let's polish it up)

vlad-demchuk (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
+      return null;
+    }
+    result[robot[key]] = key;
```
polosanya, line 16: `robot[key]` is used twice, so it's better to create a variable for this value to make your code cleaner

polosanya (changes requested):
Almost perfect!

qoa11a (changes requested):
Let's rename result variable how it was asked in previous comments.

PR #1727, opened 2022-12-14
solution problem

Ilya-Osoben:


src/inverseRobot.js
```diff
@@ -7,7 +7,33 @@
...
+  for (let key in robot) {
+    const savedKey = key;
+    const savedValue = robot[key];
+
+    delete robot[key];
+    key = savedValue;
+    robot[key] = savedKey;
+  }
+
+  return robot;
```
tetiana-ivanchenko, line 36: you can just create new object before creating loop, and in the loop you can check - if value already exist in your new object - return null, if not - create it.
In helps you avoid difficult algorithm and redundant nested loop

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
OleziO: Not resolved, if you have any questions - ask them in `fe_chat`

OleziO (changes requested):

PR #1730, opened 2022-12-16
Solution

yana-kotsulym:


src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+    newRobot[type] = robotKeys[i];
+  }
+
+  for (let i = 0; i < robotValues.length; i++) {
+    for (let n = i + 1; n < robotValues.length; n++) {
+      if (robotValues[i] === robotValues[n]) {
+        newRobot = null;
+      }
+    }
+  }
```
tetiana-ivanchenko, line 26: you need just one loop, if in your newRobot already has value that you want to add - return null, if not - add it to the new obj

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+  const robotKeys = Object.keys(robot);
```
tetiana-ivanchenko, line 11: redundant

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+
+  for (let i = 0; i < robotValues.length; i++) {
+    const type = robotValues[i];
+
+    if ([type] in newRobot) {
+      return null;
+    } else {
+      newRobot[type] = robotKeys[i];
+    }
+  }
```
tetiana-ivanchenko, line 22:
```suggestion
  for (const key in robot) {
    if (robot[key] in newRobot) {
      return null;
    }
    
    newRobot[robot[key]] = key;
  }
```

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1732, opened 2022-12-18
solution

Malva37:


nzrii-kysel (approved):
gj

src/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
```
alexandra-protyanova, line 2: odd file

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+    if (newKey in newObj) {
+      return null;
+    }
+
+    newObj[newKey] = key;
+  }
+
+  return newObj;
 }
```
alexandra-protyanova, line 23:
```suggestion
  for (const key in robot) {
    if (newKey in newObj) {
      return null;
    }

    newObj[newKey] = key;
  }
  
  const newKey = robot[key];

  return newObj;
}
```

alexandra-protyanova: It is better to use `hasOwnPrototype`, because it checks in object itself

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1735, opened 2022-12-19
Solution added

andrieieva-yuliia:


OlhaYevstifieieva (approved):
Nice job!👐🏻

src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const arrKeys = Object.keys(robot);
+  const arrValues = Object.values(robot);
+
+  for (let i = 0; i < arrKeys.length; i++) {
```
Denchepornyuk, line 14: instead of an array, you can create loop for the object and remove `arrKeys` and `arrValues` variables

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1736, opened 2022-12-19
finished

barantarasNEW:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result[robot[key]]) {
+      return null;
+    } else {
+      result[robot[key]] = key;
+    }
```
OlEzhA0, line 17: when you return something in `if` statement you might not use else

```suggestion
    if (result[robot[key]]) {
      return null;
    }

    result[robot[key]] = key;

```

OlEzhA0 (approved):
GJ!

src/inverseRobot.js
barantarasNEW: Forgot, sorry.

PR #1737, opened 2022-12-21
Add Solution

OlhaYevstifieieva:


andrieieva-yuliia (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const resultRobot = {};
+  let resultValue;
```
alexandra-protyanova, line 11: Doesn't make sense to create this variable before loop, as you don't use it after

alexandra-protyanova (approved):

PR #1738, opened 2022-12-21
Solution

anna-inozemets:
Function inverseRobot was realised

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -7,7 +7,26 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const arrKeys = Object.keys(robot);
+  const arrValues = Object.values(robot);
+  const uniqueValues = [];
+
+  for (const char of arrValues) {
+    if (!uniqueValues.includes(char)) {
+      uniqueValues.push(char);
+    }
+  }
+
+  if (uniqueValues.length !== arrValues.length) {
+    return null;
+  }
+
+  for (let i = 0; i < arrKeys.length; i++) {
+    newRobot[arrValues[i]] = arrKeys[i];
+  }
+
+  return newRobot;
 }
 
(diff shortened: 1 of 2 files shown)


src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+    if (!uniqueValues.includes(char)) {
+      uniqueValues.push(char);
+    }
+  }
+
+  if (uniqueValues.length !== arrValues.length) {
+    return null;
+  }
+
+  for (let i = 0; i < arrKeys.length; i++) {
```
BrandonSunrise, line 25: As for me It's too complicated.
You can do it with a single loop only.

BrandonSunrise (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+  const newRobot = {};
+  const arrKeys = Object.keys(robot);
+  const arrValues = Object.values(robot);
+  const uniqueValues = [];
+
+  for (const char of arrValues) {
+    if (!uniqueValues.includes(char)) {
+      uniqueValues.push(char);
+    }
+  }
```
Denchepornyuk, line 19: you don't need logic with arrays, you can create loop for the object `robot` and do everything only with newRobot and robot objects

Denchepornyuk (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    newRobot[robot[key]] = key;
```
Denchepornyuk, line 17: create a variable for `robot[key]`

Denchepornyuk (approved):

PR #1746, opened 2022-12-24
Develop: Created inverseRobot function, achieved solution

sonyaxent:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+  for (const value of Object.values(robot)) {
+    if (counts[value]) {
+      return null;
+    } else {
+      counts[value] = 1;
+    }
+  }
+
+  return inverse;
 }
```
SanyaBratashchuk, line 27:
```suggestion
  for (const key in robot) {
     const value = robot[key]
      if (value in inverse) {
        return null;
      }
    inverse[value] = key;
  }
  
  return inverse;
}
```

SanyaBratashchuk (changes requested):

agniya-i (approved):

PR #1747, opened 2022-12-24
add solution

Anastasiia-Khudych:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+  for (const elem of values) {
+    if (values.indexOf(elem) !== values.lastIndexOf(elem)) {
+      return null;
+    }
+  }
+
+  const resultObj = {};
+
+  for (let i = 0; i < values.length; i++) {
+    resultObj[values[i]] = keys[i];
```
vladyslav78292, line 22: i think you can write this row in the first loop and delete the second one. but before that move   const resultObj = {};
up

vladyslav78292 (approved):

src/inverseRobot.js
Anastasiia-Khudych: That makes sense, thanks!

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+  const keys = Object.keys(robot);
+  const resultObj = {};
+
+  for (let i = 0; i < values.length; i++) {
+    if (values.indexOf(values[i]) !== values.lastIndexOf(values[i])) {
```
alexandra-protyanova, line 15: It works, but it is gonna be more readable  to check, if your `resultObj` has this value as a key, if yes return null

alexandra-protyanova (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+  const keys = Object.keys(robot);
+  const resultObj = {};
+
+  for (let i = 0; i < values.length; i++) {
```
alexandra-protyanova, line 14: `for of` loop is gonna be more suitable here

alexandra-protyanova (approved):
GJ!

PR #1752, opened 2022-12-26
Solution

Ch0p3r84:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const valuesCheck = Object.values(robot);
+
+  for (let i = 0; i < valuesCheck.length - 1; i++) {
+    for (let j = i + 1; j < valuesCheck.length; j++) {
+      if (valuesCheck[i] === valuesCheck[j]) {
```
polosanya, line 14: This solution looks a bit complicated. Nested loops are not the best idea.
Think about creating an empty object for a new robot. Then you iterate the keys of the robot that you receive as a parameter. If this key is present in new robot - return null. If not, create an inversed property in a new robot. 
That's it

polosanya (changes requested):
Good job! But your solution is not optimized. Please check my comment and try to implement it

polosanya (approved):
Well done!

PR #1753, opened 2022-12-26
Solution

oleh-holovnykh:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+  const invertedCopy = {};
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotValues.length; i++) {
+    for (let j = i + 1; j < robotValues.length; j++) {
+      if (robotValues[i] === robotValues[j]) {
+        return null;
+      }
+    }
+  }
```
GarryRocksman, line 19: have you checked for absent property here? It's a lot of code for it.  Just use  - .hasOwnProperty() and it will take you two lines of code/

src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+
+  for (const key in robot) {
+    invertedCopy[robot[key]] = String(key);
```
GarryRocksman, line 22:
```suggestion
    invertedCopy[robot[key]] = key;
```

GarryRocksman: and use a variable for `robot[key]`

GarryRocksman (changes requested):
GJ, But need some fixes

src/inverseRobot.js
oleh-holovnykh: agree. thank you

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotValues.length - 1; i++) {
+    if (robotValues.includes(robotValues[i], i + 1)) {
+      return null;
+    }
+  }
```
Denchepornyuk, line 17: you don't need `robotValues` array for this task
you can check if key exist in object with `hasOwnProperty` method

Denchepornyuk (changes requested):

src/inverseRobot.js
oleh-holovnykh: but i need to check "If any of the object values are repeated, return 'null'."
For this i use `robotValues ` and loop

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (invertedRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    invertedRobot[robot[key]] = key;
```
Denchepornyuk, line 17: create variable for `robot[key]`

Denchepornyuk (approved):

PR #1754, opened 2022-12-26
solution

imariash86:


src/inverseRobot.js
```diff
@@ -6,8 +6,36 @@
  * @return {object}
 */
 
+/*
 function inverseRobot(robot) {
```
ann-yurkevych, line 10: code in github should be clean and without comments) just delete this)

ann-yurkevych (approved):

src/inverseRobot.js
imariash86: > code in github should be clean and without comments) just delete this)

thanks for review, I just didn't notice that:)

alexandra-protyanova (approved):

PR #1755, opened 2022-12-26
Solution

sv-dubov:


nzrii-kysel (approved):
gj👍

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const key in robot) {
+    if (robot[key] in inversed) {
```
nzrii-kysel, line 13: I'd use a safer way to check object properties

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+      return null;
+    }
+    inversed[robot[key]] = key;
```
Denchepornyuk, line 16:
```suggestion
    if (robot[key] in inversed) {
      return null;
    }

    inversed[robot[key]] = key;
```

Denchepornyuk, line 16: create variable for `robot[key]`, repeating code should be in the variables

Denchepornyuk (changes requested):

Denchepornyuk (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const key in robot) {
+    const robotKey = robot[key];
+
+    if (robotKey in inversed) {
```
Denchepornyuk, line 15: it's better to use hasOwnProperty instead of `in`

sv-dubov: Done!

PR #1762, opened 2022-12-27
inverseRobot solution

SamirMamedov1010:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const values = Object.values(robot);
+
+  for (let i = 0; i < values.length; i++) {
```
agniya-i, line 13: let's make it more simple, 
remove this loop

src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+    if (values
+      .slice(i + 1, values.length)
+      .includes(values[i])) {
+      return null;
+    }
+  }
+
+  for (const key in robot) {
```
agniya-i, line 21:
```suggestion
  for (const key in robot) {
     if (robot[key] in result) {
      return null;
    }
```
you can add this check right on `for...in` loop

agniya-i (changes requested):

ViktorHavryliuk (approved):

PR #1763, opened 2022-12-27
Solution

chermak-oleh:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const key in robot) {
+    if (inversed.hasOwnProperty(robot[key])) {
+      return null;
+    }
```
Anastasiia-Khudych, line 15: You should use `Object.prototype.hasOwnProperty()` instead of just `.hasOwnProperty`

Anastasiia-Khudych (changes requested):

src/inverseRobot.js
chermak-oleh: Sorry, but i did not understand what do u mean.  Look at the screenshot from checklist: https://prnt.sc/LOFeMFoO9cwc

Anastasiia-Khudych: I've read about it in the style-guide. 
image
here https://github.com/mate-academy/style-guides/blob/master/javascript.md#37-do-not-callobjectprototypemethods-directly-such-ashasownpropertypropertyisenumerable-andisprototypeof

Denchepornyuk, line 15: create variable for `robot[key]`

Denchepornyuk (approved):

PR #1764, opened 2022-12-27
Solution

karaedd:


src/inverseRobot.js
```diff
@@ -7,7 +7,12 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverse = Object.fromEntries(Object.entries(robot).map((
+    [key, value]) => [value, key]));
```
Bohdan-Maksymiuk, line 11:
```suggestion
  const inverse = Object.fromEntries(Object.entries(robot)
    .map(([key, value]) => [value, key]));
```

src/inverseRobot.js
```diff
@@ -7,7 +7,12 @@
...
+
+  return Object.keys(
+    inverse).length < Object.values(
+    robot).length ? null : inverse;
```
Bohdan-Maksymiuk, line 15:
```suggestion
  return Object.keys(inverse).length < Object.values(robot).length
    ? null
    : inverse;
```

Bohdan-Maksymiuk (changes requested):
I left general code style comments
but also you can improve complexity of your code:
- create empty object for inversed robot
- iterate once by initial object and fulfill reversed robot
- at the start of iteration check if the key exists (our duplicate) and return from fucntion

Bohdan-Maksymiuk (approved):
Great job 👍

PR #1765, opened 2022-12-27
created function

ann-yurkevych:


andrieieva-yuliia (approved):

src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
...
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
+
+    if (newKey in inversedRobot) {
+      return null;
+    }
+    inversedRobot[newKey] = key;
+  }
```
alexandra-protyanova, line 20:
```suggestion
  for (const key in robot) {
    if (newKey in inversedRobot) {
      return null;
    }
  
    const newKey = robot[key];
    
    inversedRobot[newKey] = key;
  }
```

alexandra-protyanova: It is pointless to create variable if you will return null)

alexandra-protyanova (approved):
GJ!

PR #1766, opened 2022-12-27
solution

yaroslavna-horban:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (robot[key] in result) {
```
ann-yurkevych, line 13: you can add variable for robot[key], but it's additional)

ann-yurkevych (approved):
Good job)) Excellent!

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+      return null;
+    }
+
+    result[robot[key]] = key;
```
Denchepornyuk, line 17: create variable for `robot[key]`

src/inverseRobot.js
Denchepornyuk, line 13: it's better to use `hasOwnProperty` instead of `in`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1767, opened 2022-12-27
Task was solved

ihor-soloviov:


src/inverseRobot.js
```diff
@@ -8,6 +8,20 @@
...
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+  const robotValues = Object.values(robot);
+
+  for (const char of robotValues) {
+    if (robotValues.indexOf(char) !== robotValues.lastIndexOf(char)) {
+      return null;
+    }
+  }
```
vanst1, line 18: redundant code, you can to it more simple

src/inverseRobot.js
```diff
@@ -8,6 +8,20 @@
...
+
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
```
vanst1, line 22:
```suggestion
  for (const key in robot) {
  if (newRobot.hasOwnProperty(robot[key])) {
      return null;
    }
    newRobot[robot[key]] = key;
  }
```

stadnytskyi-ivan (changes requested):

src/inverseRobot.js
ihor-soloviov: Yap. Thanks!

stadnytskyi-ivan (approved):

alexandra-protyanova (approved):

PR #1768, opened 2022-12-27
solution

Dashaverz:


nzrii-kysel (approved):
clean, well done

src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
```
alexandra-protyanova, line 14: Make sense to create this variable after first return as in case of return you will not need it

alexandra-protyanova (approved):
Well done 👍

PR #1774, opened 2022-12-28
Result

svitjojo:


andrieieva-yuliia (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    const prop = robot[key];
```
alexandra-protyanova, line 13: Make sense to create this variable after first return as in case of return you will not need it

alexandra-protyanova (approved):
Well done 👍

PR #1775, opened 2022-12-28
solution

OlehSharha:


src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverse = {};
```
victorionov, line 10: Its better to make more proper name for new object. 
everithing else is perfect.

victorionov (approved):

src/inverseRobot.js
alexandra-protyanova: As for me, not so bad name for variable) But `inverseRobot` looks better

alexandra-protyanova (approved):

PR #1776, opened 2022-12-28
Solution

Coyoote:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invRobot = {};
+
+  for (const key in robot) {
+    if (invRobot.hasOwnProperty(robot[key])) {
```
andrieieva-yuliia, line 13: need create a variable for robot[key] 🙂

andrieieva-yuliia (changes requested):

src/inverseRobot.js
Coyoote: Added the variable, thank you for the review 😄

Denchepornyuk (approved):

PR #1777, opened 2022-12-28
Added Solution

burovetskyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const values in robot) {
+    const newValue = robot[values];
```
ann-yurkevych, line 13: you can create variable const newValue = robot [values] after return null, it'd be more appropriately

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+    if (newRobot.hasOwnProperty(newValue)) {
+      return null;
+    }
+
```
ann-yurkevych, line 18:
```suggestion
  for (const values in robot) {

    if (newRobot.hasOwnProperty(newValue)) {
      return null;
    }
    const newValue = robot[values];

```

ann-yurkevych (approved):

Denchepornyuk (approved):

PR #1780, opened 2022-12-28
v-01

AndreyBerezyanskiy:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const rightRobot = {};
+  const valueArr = Object.values(robot);
+  const sortArr = valueArr.sort();
+
```
andrii-kurash, line 13: you can rewrite method `sort` like this: `sort((a, b) => a - b)`. Its better practice, because method sort can work incorrect
It sort array only by the first number, for example [1, 200, 40, 5].  It is better to specify the sort order directly. That's what Denis said!)

andrii-kurash (approved):
Well done!👍)

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const rightRobot = {};
+  const valueArr = Object.values(robot);
+  const sortArr = valueArr.sort();
```
Denchepornyuk, line 12: you don't need array for this task, you can make loop for the object and make check for repeated keys in this loop (use hasOwnProperty for this)

Denchepornyuk (changes requested):

AndreyBerezyanskiy:
Thanks!🙂

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const rightRobot = {};
+  // const valueArr = Object.values(robot);
```
Denchepornyuk, line 11: remove comments

Denchepornyuk (approved):

PR #1784, opened 2022-12-28
Solution

maasiimka:


src/inverseRobot.js
```diff
@@ -8,6 +8,21 @@
 
 function inverseRobot(robot) {
   // write code here
+  const valueArr = Object.values(robot);
+
+  for (let i = 0; i < valueArr.length; i++) {
+    if (valueArr.indexOf(valueArr[i]) !== valueArr.lastIndexOf(valueArr[i])) {
+      return null;
+    }
+  }
```
sv-dubov, line 17: Maybe, its better to use hasOwnProperty()

sv-dubov (approved):
Interesting solution

src/inverseRobot.js
maasiimka: Yes, it seems quite better than my solution. Thanks!

ArturErmolenko (approved):

PR #1788, opened 2022-12-28
Develop

victorionov:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    if (reversedRobot.hasOwnProperty(robot[key])) {
```
andrieieva-yuliia, line 13: You should create a variable for robot[key] 🙂

andrieieva-yuliia (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    const trueKey = robot[key];
```
Denchepornyuk, line 13: `trueKey` isn't a good name for a variable, why true?

Denchepornyuk (approved):

PR #1792, opened 2022-12-28
solution

Oleksandr-Maly:


AndreyBerezyanskiy (approved):
Nice job!👍

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotNew = {};
+
+  for (const keys in robot) {
```
alexandra-protyanova, line 12:
```suggestion
  for (const key in robot) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+    robotNew[robot[keys]] = keys;
+  }
+
+  if (Object.keys(robotNew).length !== Object.values(robot).length) {
```
alexandra-protyanova, line 16:
```suggestion
  if (Object.keys(robotNew).length !== Object.keys(robot).length) {
```

alexandra-protyanova: Sense is the same, but more understandable as for me

alexandra-protyanova (approved):

PR #1799, opened 2022-12-29
solution

Alessandranitti15:


src/inverseRobot.js
```diff
@@ -7,7 +7,30 @@
...
+  // const keysFrom = Object.entries(robot); array dall'oggetto
+
+  // for (let i = 0; i < keysFrom.length; i++) {
+  //   keysFrom[i].reverse(keysFrom);
+  // } per invertire ogni array dentro il nuovo array keysfrom
+
+  // const newObject = Object.fromEntries(keysFrom);
+  // riconvertire tutto in oggetto
+
+  // return newObject;
```
qoa11a, line 33: Remove commented code.

src/inverseRobot.js
```diff
@@ -7,7 +7,30 @@
...
+  const robotInverso = {};
+
+  for (const key in robot) {
+    const robotKey = robot[key];
+
+    if (robotInverso.hasOwnProperty(robotKey)) {
+      return null;
+    }
+
+    robotInverso[robot[key]] = key;
```
qoa11a, line 19: You already have variable for robot[key] - use it.

src/inverseRobot.js
```diff
@@ -7,7 +7,30 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInverso = {};
+
+  for (const key in robot) {
+    const robotKey = robot[key];
```
qoa11a, line 13: When you get key in object the result is value. So it's a little bit confusing that variable name is robotKey.

qoa11a (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInverso = {};
+
+  for (const value in robot) {
+    const robotValue = robot[value];
```
polosanya, line 13:
```suggestion
  for (const property in robot) {
    const robotValue = robot[property];
```
Actually, [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) iterates properties(keys) of object, so previous comment was only about changing robotKey to robotValue

polosanya (approved):
Great!

PR #1800, opened 2022-12-29
solution

Anton-Zhytonbaiev:


oleh-holovnykh (approved):
Nice!

alexandra-protyanova (changes requested):
Both loops, work in the same way, combine them. And you can use `for in` to iterate through object

alexandra-protyanova (approved):

PR #1802, opened 2022-12-29
solution_1

Borys-Andrew:


src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
...
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]] !== undefined) {
+      return null;
+    }
+
+    newRobot[robot[key]] = key;
```
Coyoote, line 18: You need to create a variable for robot[key]

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]] !== undefined) {
```
Coyoote, line 14: You're checking if the value of newRobot key isn't empty, instead you should compare if the newRobot object doesn't already have a similar key

Coyoote (changes requested):
Good job, left a comment regarding the if statement

src/inverseRobot.js
Borys-Andrew: in this case i agree with you, and new if statement   `if (newRobot[robot[key]]) ` will check if any of the input object values are repeated in output object.

src/inverseRobot.js
Denchepornyuk: agree

denys-cheporniuk (approved):

PR #1803, opened 2022-12-29
inverse robot solution

serbinm:
May be a bit complicated, but I love iterating :3

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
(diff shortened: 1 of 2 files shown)


sv-dubov (approved):
Interesting solution

ArturErmolenko (changes requested):
nested loops is EVIL, your code has bad performance , complexity O(n), so you could do this task using only one `for in` loop 
You could watch about algorithms complexity here
[link](https://mate.academy/learn/algorithms/algorithm-complexity#/video/660)

ArturErmolenko (approved):

PR #1809, opened 2023-01-05
Finish task

U-Ra-All:


src/inverseRobot.js
```diff
@@ -4,10 +4,28 @@
...
 function inverseRobot(robot) {
   // write code here
+  const robotKeys = [];
+  const robotValues = [];
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    robotKeys.push(key);
+    robotValues.push(robot[key]);
+  }
```
witflash, line 18:
```
const value = robot[key];

if (reversedRobot[value]) {
  return null;
}

reversedRobot[value] = key;
```

And you should not need to use robotKeys, robotValues, and additional `for` block.

witflash (changes requested):
Good job!
But try to optimize your code with my comment.

src/inverseRobot.js
U-Ra-All: Thanks, done!

witflash (approved):
Well done 👍🏻

PR #1811, opened 2023-01-05
Solution

ia-lerimund:


src/inverseRobot.js
```diff
@@ -8,6 +8,21 @@
 
 function inverseRobot(robot) {
   // write code here
+  const array = Object.values(robot);
+  const isValuesAreDuplicated = array.some(item => {
+    if (array.indexOf(item) !== array.lastIndexOf(item)) {
+      return true;
+    }
+  });
```
OlEzhA0, line 16: why do we need like this? 
Create variable newRobot, move through of your object, and on each iteration check, if your newRobot has key, if not - add key and value to newRobot

OlEzhA0 (changes requested):
Hey! check please my comment

src/inverseRobot.js
ia-lerimund: I've changed the solution, thank you for your advice!

PR #1814, opened 2023-01-07
solution inverse robot

Spec0s:


src/inverseRobot.js
```diff
@@ -8,6 +8,21 @@
...
+    inverseArr[robotValues[i]] = robotKeys[i];
+  }
+
+  const inverseArrKeys = Object.keys(inverseArr);
+
+  if (robotKeys.length !== inverseArrKeys.length) {
+    return null;
+  }
+
+  return inverseArr;
```
qoa11a, line 25: This solution works, but let's try to do it another way. You can use for .. in loop to get keys of object, than you can access its props as object[key].

qoa11a (changes requested):

src/inverseRobot.js
Spec0s: hello i also noticed you had in checklist best example 
 "EVEN BETTER EXAMPLE if (robot.hasOwnProperty(key)) {... "
 
 but i could not figure out how to use it here, can you help me ?

src/inverseRobot.js
```diff
@@ -8,6 +8,21 @@
...
+    inverseArr[robot[key]] = key;
+  }
+
+  function getObjectLength(obj) {
+    return (Object.keys(obj)).length;
+  }
+
+  if (getObjectLength(robot) !== getObjectLength(inverseArr)) {
+    return null;
+  }
```
polosanya, line 23: Do you really need this logic? 
You can just check inside the loop if key is already present in inversedArr as follow:
```suggestion
  if (inversedArr.hasOwnProperty(key)) {
    return null;
  }
```

src/inverseRobot.js
```diff
@@ -8,6 +8,21 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inverseArr = {};
```
polosanya, line 11:
```suggestion
  const inversedArr = {};
```
or even better - inversedRobot

polosanya (changes requested):
Almost perfect!

src/inverseRobot.js
Spec0s: i feel so stupid right now,.. i was on this so many hours and the solution was so obvious...  :( thank you sensei !

polosanya (approved):
Well done!)

PR #1817, opened 2023-01-09
js / situs inversus

valntyn:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key of Object.keys(robot)) {
+    const a = robot[key];
```
tetiana-ivanchenko, line 13: use more meanfull variable names

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
valntyn: thank you for the correction
I'll keep it in mind :)

IvanFesenko (approved):

PR #1820, opened 2023-01-12
Solution

bojkovladislav:


Virass (changes requested):
Don't use loop inside loop in this task. It not good for performance and not good way for solution this task.

SanyaBratashchuk (approved):

PR #1822, opened 2023-01-13
commit message

Peti51:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
-  // write code here
+  const keys = Object.keys(robot);
+
+  for (let i = 0; i <= keys.length; i++) {
+    for (let j = i + 1; j <= keys.length; j++) {
+      if (robot[keys[i]] === robot[keys[j]]) {
+        return null;
+      }
+    }
+  }
```
polosanya, line 18: Do you really need two nested loops to check it? (and three loops at all)
I think only one iteration of robot keys is enough.
Let's try to optimize your solution.
You can just check it newRobot has properties from robot

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+
+  const result = {};
```
polosanya, line 20: image

polosanya (changes requested):
Let's polish it up)

Peti51:
This is the only way I chould solve it and have only one iteration in the code.

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+  for (const key in robot) {
+    newRobot[robot[key]] = key;
+  }
+
+  const oldRobotKeys = Object.values(robot).sort();
+  const newRobotKeys = Object.keys(newRobot).sort();
+
+  if (newRobotKeys.length !== oldRobotKeys.length) {
+    return null;
+  }
```
polosanya, line 21:
```suggestion
  for (const key in robot) {
    if (newRobot.hasOwnProperty(robot[key])) {
      return null;
    }
    
    newRobot[robot[key]] = key;
  }
```
Try this
Also, it's better to create a variable for robot[key] as it's used twice

polosanya (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,29 @@
...
+  // }
+
+  // const oldRobotKeys = Object.values(robot).sort();
+  // const newRobotKeys = Object.keys(newRobot).sort();
+
+  // if (newRobotKeys.length !== oldRobotKeys.length) {
+  //   return null;
+  // }
+
+  // return newRobot;
```
polosanya, line 23: Don't forget to delete the comments

polosanya (approved):
Well done!

PR #1825, opened 2023-01-15
task done

tecHamster:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
```
polosanya, line 10: Let's use more descriptive names for the variables. What is obj?
Try to specify what this variable is

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const key in robot) {
+    if (robot[key] in obj) {
+      return null;
+    }
+
+    obj[robot[key]] = key;
```
polosanya, line 17: You use robot[key] twice, so it's better to create a variable for that value, to make your code cleaner

polosanya (changes requested):
Let's polish it up)

qoa11a (approved):

PR #1829, opened 2023-01-16
Task complete

Sergei-Nikolaienko:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const swapped = {};
+
+  for (const key in robot) {
+    if (swapped.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      swapped[robot[key]] = key;
+    }
```
Viktor-Kost, line 17: when you have return inside if statement - no needs to use else statement
```suggestion
    } 
    
    swapped[robot[key]] = key;
    
```

Viktor-Kost (approved):
Nice!

Apply some improvement

PR #1832, opened 2023-01-17
Solution

kirill-vypirovskyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+  const entries = Object.entries(robot);
```
gretoq, line 11: Nice idea

gretoq (approved):
Good job!

src/inverseRobot.js
Andr1yK, line 11: +O(n) to complexity, nothing more)

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const entry of entries) {
```
Andr1yK, line 13: use `for in`, and then u can avoid Object.entries()

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+    if (fixedRobot.hasOwnProperty(entry[1])) {
+      return null;
+    }
+    fixedRobot[entry[1]] = entry[0];
+  }
```
Andr1yK, line 18: FYI: better to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

```suggestion
    const [key, value] = entry;

    if (fixedRobot.hasOwnProperty(value)) {
      return null;
    }

    fixedRobot[value] = key;
  }
```

Andr1yK (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const entry in robot) {
+    if (fixedRobot.hasOwnProperty([robot[entry]])) {
+      return null;
+    }
+    fixedRobot[robot[entry]] = entry;
+  }
```
Andr1yK, line 17:
```suggestion
    if (fixedRobot.hasOwnProperty([robot[entry]])) {
      return null;
    }

    fixedRobot[robot[entry]] = entry;
  }
```

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const entry in robot) {
```
Andr1yK, line 12:
```suggestion
  for (const key in robot) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+    if (fixedRobot.hasOwnProperty([robot[entry]])) {
```
Andr1yK, line 13:
```suggestion
    if (fixedRobot.hasOwnProperty(robot[entry])) {
```

Andr1yK (approved):

PR #1835, opened 2023-01-17
Solution

G37ua:


qoa11a (changes requested):
techically this solution works, but there's more optimal way to solve this task

hint: you should return null right away, when you meet existing key

G37ua:
Второй вариант

Anastasiia-Svintsova (approved):

PR #1836, opened 2023-01-17
Solution

annperetiatko:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const allValues = [];
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (allValues.includes(value)) {
```
witflash, line 16: You can check 
```
if (!newRobot.value) {
  newRobot.value = key;
}
```

So you can avoid additional variable creation and additional `else` block.

witflash (changes requested):
Well done!
Could you do small fix, please?

Viktor-Kost (approved):

PR #1839, opened 2023-01-19
Solution

Yevhen-Ivanchuk:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```

polosanya (changes requested):
Almost perfect! One small fix

Anastasiia-Svintsova (approved):

PR #1847, opened 2023-01-22
Tests were passed succsessfully

read-my-code:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
-  // write code here
+  const newRob = {};
+
+  for (const key in robot) {
+    newRob[robot[key]] = key;
+  }
+
+  if (Object.keys(newRob).length !== Object.values(robot).length) {
+    return null;
+  }
```
DarkMistyRoom, line 18: would be better to make this check in the loop to return null immediately when we find the repeated value and avoid unnecessary actions

DarkMistyRoom (changes requested):

ViktorHavryliuk (approved):

PR #1848, opened 2023-01-22
Solution

MTS13:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const ret = {};
```
ViktorHavryliuk, line 10:
```suggestion
  const result = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+
+  const values = Object.values(robot);
```
ViktorHavryliuk, line 12: Better to move it down near values of resutl declaration, as you are using this two vars together and it will be easier to follow.
However, you can omit using those at all. There is an `in` operator in JS, which lets you to check if some object has some field or not, like `if ("foo" in { foo: "bar"})`. You can use it in your existing for loop to see if result already has some key, and return null if so, for example

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):

PR #1854, opened 2023-01-24
task was solved

vanyachyzh:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const robotLength = Object.keys(robot).length;
+  let newRobotLength = 0;
```
Viktor-Kost, line 12: no needs to check and use length

Viktor-Kost (changes requested):

Viktor-Kost (approved):
Nice!

PR #1858, opened 2023-01-25
Solution

silcki:


src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
...
+
+  if (duplicateElements.length > 0) {
+    return null;
+  }
+
+  const inversedRobot = {};
+
+  Object.keys(robot).forEach((value, index) => {
+    inversedRobot[keys[index]] = value;
+  });
```
Bohdan-Maksymiuk, line 24: you can do all previous logic with this logic and iterate only ones (but you will need to rewrite it into for loop)
you check here if key existed in inversedRobot and return `null`

Bohdan-Maksymiuk (changes requested):

PR #1859, opened 2023-01-26
solution

MeryamWazeen:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseKeys = {};
```
```suggestion
  const reversedRobot = {};
```
Actually, it's not keys, but a new robot

polosanya (approved):
Good job!

PR #1862, opened 2023-01-27
solved

danulyk05:


src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+  }
+
+  for (let i = 0; i < values.length; i++) {
+    if (values.includes(values[i], i + 1)) {
+      return null;
+    }
+  }
 
+  for (const key in robot) {
+    revers[robot[key]] = key;
```
ViktorHavryliuk, line 27: You can also use in operator to check if object has some property, by doing it here you can remove two loops above
Also you may check it via reverse[robot[key]] === undefined or revers.hasOwnProperty(robot[key])
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
```suggestion
    if(robot[key] in revers) {
      return null
    }

    revers[robot[key]] = key;
```

src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+    if (keys.includes(keys[i], i + 1)) {
+      return null;
+    }
+  }
+
+  for (let i = 0; i < values.length; i++) {
+    if (values.includes(values[i], i + 1)) {
+      return null;
+    }
+  }
```
ViktorHavryliuk, line 24:
```suggestion
```

ViktorHavryliuk (changes requested):

src/inverseRobot.js
danulyk05: Thank you very much

ViktorHavryliuk (approved):

PR #1863, opened 2023-01-27
Solution

MarinaOlenchenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversed = { ...robot };
```
ViktorHavryliuk, line 10:
```suggestion
  const reversed = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+
+  for (const key of Object.keys(reversed)) {
+    if (reversed.hasOwnProperty(reversed[key])) {
+      return null;
+    }
+
+    reversed[reversed[key]] = key;
+    delete reversed[key];
```
ViktorHavryliuk, line 18:
```suggestion
```

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const reversed = { ...robot };
+
+  for (const key of Object.keys(reversed)) {
+    if (reversed.hasOwnProperty(reversed[key])) {
+      return null;
+    }
+
+    reversed[reversed[key]] = key;
```
ViktorHavryliuk, line 17: You can also use for...in (remember that it's different from for...of) loop to iterate robot's keys: `for (const key in robot)`, and use in operator to check if object has property: `if (robot[key] in reversed)`
But that's up to you as it is more of matter of taste

```suggestion
  for (const key of Object.keys(robot)) {
    if (reversed.hasOwnProperty(robot[key])) {
      return null;
    }

    reversed[reversed[key]] = key;
```

ViktorHavryliuk (changes requested):
You can reduce amount of code and avoid keys deletion if you start with empty result object instead of copying robot

PR #1878, opened 2023-01-30
done

Kocha1991:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (value in newObj) {
+      return null;
+    }
+    newObj[value] = key;
+  }
```
loralevitska, line 17:
```suggestion
  for (const [key, value] of Object.entries(robot)) {
    if (value in newObj) {
      return null;
    }

    newObj[value] = key;
  }
```

loralevitska (changes requested):
Almost perfect) Let's improve code style

ViktorHavryliuk (approved):

PR #1884, opened 2023-02-02
Solution

Zavalishyn-Oleksandr:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const revert = {};
+  const exclusiveNames = [];
+
+  for (const key in robot) {
+    if (exclusiveNames.includes(robot[key])) {
```
tetiana-ivanchenko, line 14:
```suggestion
    if (robot[key] in revert) {
```

tetiana-ivanchenko: in this case your exclusiveNames is redundant

tetiana-ivanchenko (changes requested):

Zavalishyn-Oleksandr:
Thank you, now it seems much better

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const revert = {};
+
+  for (const key in robot) {
+    if (robot[key] in revert) {
+      return null;
+    }
+    revert[robot[key]] = key;
```
ViktorHavryliuk, line 16: Would be good to add empty line here for better readability. In this case it's ok without it, but it’s a good habit to have when code becomes bigger
```suggestion
    if (robot[key] in revert) {
      return null;
    }
    
    revert[robot[key]] = key;
```

ViktorHavryliuk (approved):
Good work

PR #1887, opened 2023-02-03
inverse_robot - first commit

RufatRM:


src/inverseRobot.js
```diff
@@ -7,7 +7,31 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotValues.length; i++) {
+    for (let j = 0; j < robotValues.length; j++) {
```
polosanya, line 13: It is a bad practice to use nesting loops. Think about making all the needed action in one loop

src/inverseRobot.js
```diff
@@ -7,7 +7,31 @@
...
+  const robotEntries = Object.entries(robot);
+  const objectReversed = {};
+
+  for (let i = 0; i < robotEntries.length; i++) {
+    const innerArray = robotEntries[i].reverse();
+
+    const objectKey = innerArray[0];
+    const objectValue = innerArray[1];
+
+    objectReversed[objectKey] = objectValue;
```
polosanya, line 31: Can you do all the checks in this loop?

polosanya (changes requested):
Good job! But let's try to improve your code. Think about using only one loop in this task to improve the performance

src/inverseRobot.js
RufatRM: hi, I searched on the internet about how to check if an array contains duplicate elements, and the most easiest and understandable way for me to solve it was javascript Set object, so I came up with that solution.

polosanya (changes requested):
You can just iterate through all keys of robot and add them reversed to the new reversed robot.
You only need to check if property is already present in reversed robot and return null in that case
  ```
for (const key in robot) {
    if (robot[key] in reversedRobot) {
      return null;
    }

    reversedRobot[robot[key]] = key;
  }
```

polosanya (approved):
Well done!

PR #1891, opened 2023-02-03
Solution

ArtemVelychko:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
```
```suggestion
  const inversedRobot = {};
```
Let's specify the name

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const key in robot) {
+    if (inverse[robot[key]]) {
+      return null;
+    } else {
+      inverse[robot[key]] = key;
+    }
```
polosanya, line 17:
```suggestion
    if (inverse[robot[key]]) {
      return null;
    }
    
    inverse[robot[key]] = key;
```
There is no need for `else` as your code will not go further after return anyway

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments)

PR #1893, opened 2023-02-04
Solution

Nataliia95254:


src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+      }
+    }
+
+    for (const key in robot) {
+      inverseRob[robot[key]] = key;
+    }
+
+    return inverseRob;
+  }
 }
```
nazarmatsevych, line 30: Looks too complicated, you can check
if this is (inverseRob[robot[key]]) true then return null, otherwise inverseRob[robot[key]] = key;

nazarmatsevych (changes requested):
Good work,
check my comment pls

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverseRob = {};
+  const values = Object.values(robot);
+
+  if (values.length === 0) {
+    return inverseRob;
+  }
```
tetiana-ivanchenko, line 15:
```suggestion
```

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const inverseRob = {};
+
+  for (const key in robot) {
+    if (inverseRob[robot[key]]) {
```
sTorba24, line 13:
```suggestion
    if (inverseRob.hasOwnProperty(robot[key])) {
```

sTorba24: `hasOwnProperty` is better practice to check obj property

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+      return null;
+    } else {
+      inverseRob[robot[key]] = key;
+    }
```
sTorba24, line 17:
```suggestion
    } 
    inverseRob[robot[key]] = key;
```

sTorba24: you don't need else here :)

if your condition will work - it'll just return null and program stop
if skip - it will do what you want without else :)

sTorba24 (approved):
Good work! Let's improve it :)

PR #1894, opened 2023-02-06
solution

ivanm2706:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const arrRobot = Object.entries(robot);
+  const arrValues = Object.values(robot);
+
+  if ([...new Set(arrValues)].length !== arrValues.length) {
```
mykhalenych, line 13: looks strange
better use for in circle
for (const key in robot) => check if robot[key] !== null else add to object with key robot[key] value key

mykhalenych (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const resultRobot = {};
+
+  for (const key in robot) {
+    for (const keyResultRobot in resultRobot) {
+      if (robot[key] === keyResultRobot) {
+        return null;
+      }
+    }
```
tetiana-ivanchenko, line 17:
```suggestion
  for (const key in robot) {
    if (robot[key] in resultRobot) {
      return null;
    }  
```

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const resultRobot = {};
+
+  for (const key in robot) {
+    if(resultRobot.hasOwnProperty(robot[key])){
```
tetiana-ivanchenko, line 13:
```suggestion
    if (resultRobot.hasOwnProperty(robot[key])) {
```

tetiana-ivanchenko (approved):

PR #1896, opened 2023-02-06
inverseRobot

IvanFront-end:


IvanFront-end:
Hi! Can you give the second solution? maybe there is an easier method?

nazarmatsevych (approved):
Your solution is fine, but this is how I would complete the task
`function inverseRobot(robot) {
  const inverse = {};

  for (const key in robot) {
    if (inverse[robot[key]]) {
      return null;
    }
    inverse[robot[key]] = key;
  }

  return inverse;
}`

So we created a new empty object, check if object value is not repeated, if it is then we return null, otherwise we  write the keys and values ​​in our object in a different order

PR #1902, opened 2023-02-08
Task done

sergik94:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
-  // write code here
+  const newObj = {};
+
+  for (const prop in robot) {
+    if (!newObj[robot[prop]]) {
+      newObj[robot[prop]] = prop;
+      continue;
+    }
+
+    return null;
```
etojeDenys, line 18: it`s better to rewrite
```suggestion
    if (newObj[robot[prop]]) {
      return null;
    }
    newObj[robot[prop]] = prop;
    
```

etojeDenys (changes requested):
almost done

etojeDenys (approved):
LGTM

PR #1909, opened 2023-02-09
solution

milenadiamonds1:


Yekatery (approved):
Well done!

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverse = Object.entries(robot).map(([value, key]) => [key, value]);
+
+  const flatted = reverse.flat();
+  const uniqueParts = flatted
+    .filter((element, index, array) => array.indexOf(element)
+    === index);
```
illiarerun, line 15: you can use Set for it
```suggestion
  const uniqueParts = new Set(...reverse);
```

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+
+  if (flatted.length !== uniqueParts.length) {
+    return null;
+  } else {
+    return Object.fromEntries(reverse);
+  }
```
illiarerun, line 21: [checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) 4th item

illiarerun (changes requested):

src/inverseRobot.js
milenadiamonds1: i didn't get how to use Set instead of my solution so i prefer to leave my option with filter 😅

VadKol (approved):
bon travail)

illiarerun (approved):

src/inverseRobot.js
illiarerun: You can try to do it 😀 Set contains ONLY unique values

PR #1913, opened 2023-02-10
Solution

vshepel:


vshepel:
I see that the tests do not pass, but the Object.hasOwn function that the node (v12) does not see in the tests is too modern for it, but this does not mean that I am wrong, right?

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotNormal = {};
+
+  for (const key of Object.keys(robot)) {
+    if (Object.hasOwn(robotNormal, robot[key])) {
```
etojeDenys, line 13:
```suggestion
    if (robotNormal[robot[key]]) {
```

etojeDenys (approved):
LGTM

small nit

PR #1914, opened 2023-02-10
Solution

toljan1:


src/inverseRobot.js
```diff
@@ -7,7 +7,28 @@
...
+
+  for (let i = 0; i < keys.length; i++) {
+    for (let j = i + 1; j < keys.length; j++) {
+      if (keys[i] === keys[j]) {
+        return null;
+      } else {
+        let z = 0;
+
+        for (; z < keys.length; z++) {
+          const key = keys[z];
```
etojeDenys, line 22: you don`t need to write nested loops

```
     return null;
    } else {
      robotNormal[robot[key]] = key;
    }
  }
```

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #1916, opened 2023-02-12
Solution

IvanVaverchak:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+  const rightRobot = {};
+
+  for (let i = 0; i <= realKeys.length; i++) {
+    if (rightRobot[realKeys[i]]) {
+      return null;
+    }
+    rightRobot[realKeys[i]] = realValues[i];
+  }
+
+  return rightRobot;
```
etojeDenys, line 21: we can rewrite much more simple

```
     return null;
    } else {
      robotNormal[robot[key]] = key;
    }
  }
```

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #1921, opened 2023-02-12
solution

KuharchukDmytro:


Yevhenii0536 (approved):
As for me everything is okay. But i think you may set name for new object mor infofull

src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
```
Andr1yK, line 10: cmnt

Andr1yK: super informative comment

src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
...
+
+  for (const part in robot) {
+    obj[robot[part]] = part;
+  }
+
+  return Object.values(obj).length < Object.values(robot).length
+    ? null
+    : obj;
```
Andr1yK, line 18: mb, we can check it(and return null) inside `for` loop?

Andr1yK (changes requested):

Andr1yK (approved):

PR #1924, opened 2023-02-13
Solution

timoxa2t:


NikitaStambul (approved):
good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverdedRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
```
Andr1yK, line 12: why do u need to use Object.entries here?

Avoid increasing complexity))

Andr1yK (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverdedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
```
Andr1yK, line 13: 🔥

Andr1yK (approved):

PR #1925, opened 2023-02-13
Inverse Robot - Solution

Yevhenii0536:


src/inverseRobot.js
```diff
@@ -4,10 +4,27 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
+  const arrKey = [];
+  const arrVal = [];
+
+  for (const key in robot) {
+    arrKey.unshift(key);
+    arrVal.unshift(robot[key]);
+  }
```
stanislavkulakovskyi, line 17: I guess you can optimize it a bit by assigning arrKey to Object.keys(robot) and arrVal to Object.values(robot)
that way you wouldn't need to run this loop :)

stanislavkulakovskyi (approved):
Good job! Solution could be optimized a bit, but since it works - I consider the task is done

Andr1yK (changes requested):
Wow... not the best solution (it takes a lot of resources for each call)

let's simplify))

src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
  * @param {object} robot
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const copy = {};
```
Andr1yK, line 10:
```suggestion
  const inversedRobot = {};
```

Andr1yK (approved):

PR #1929, opened 2023-02-13
inverseRobot solution

maksym-maliuk:


Lilu580 (commented):
Gj

Stepan-TS (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const key = Object.keys(robot);
+  const value = Object.values(robot);
```
Andr1yK, line 12: array with keys named as a key.. something is wrong

Andr1yK (changes requested):
now your code had a big algorithmic complexity, let's simplify it)

U can use just `for in` loop

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]]) {
+      return null;
+    }
+    newRobot[robot[key]] = key;
```
Andr1yK, line 16:
```suggestion
    if (newRobot[robot[key]]) {
      return null;
    }

    newRobot[robot[key]] = key;
```

Andr1yK (approved):

PR #1930, opened 2023-02-13
Solution

BozhenaBodnar:


MarkBabich (approved):
Works as it should, cool :)

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const correctRobor = {};
```
illiarerun, line 10: robor?

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const key in robot) {
+    if (correctRobor.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    correctRobor[robot[key]] = key;
```

illiarerun (approved):

src/inverseRobot.js
BozhenaBodnar: ok:)

PR #1933, opened 2023-02-13
sol

Lilu580:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = [];
+  const obj = {};
+
+  for (const key in robot) {
+    if (values.includes(robot[key])) {
```
sergey-lipikhin, line 14: Do not create `values` array at all and check for repeated keys with `in` operator or `hasOwnProperty` method directly in your `obj`, instead of `includes`

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+      return null;
+    } else {
+      values.push(robot[key]);
+      obj[robot[key]] = key;
+    }
```
sergey-lipikhin, line 19: **Checklist p.4**

image

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = [];
+  const obj = {};
```
sergey-lipikhin, line 11: `obj `is definitely not a good choice for variable naming

sergey-lipikhin (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const resObj = {};
```
Andr1yK, line 10: point 4 https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const key in robot) {
+    if (resObj.hasOwnProperty(robot[key])) {
```
Andr1yK, line 13: 🔥

Andr1yK (changes requested):

PR #1938, opened 2023-02-13
solution

Nordexz:


src/inverseRobot.js
```diff
@@ -8,6 +8,26 @@
...
+
+  for (let i = 0; i < values.length; i++) {
+    repaired[keys[i]] = values[i];
+  }
+
+  const check = [...new Set(keys)];
+
+  if (check.length !== keys.length) {
+    return null;
+  }
```
illiarerun, line 28: why don't you do it before filling up new object?

illiarerun (approved):

PR #1939, opened 2023-02-13
Solution

sasha-dranuta:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const reversedObject = {};
+
+  for (const key in robot) {
+    if (robot[key] in reversedObject) {
+      return null;
+    }
+
+    reversedObject[robot[key]] = key;
```

illiarerun (approved):

PR #1940, opened 2023-02-13
Solution

dudar-y:


src/inverseRobot.js
```diff
@@ -7,7 +7,30 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+  let count = 0;
```
Viktor-Kost, line 13: it's redundant
```suggestion
```

src/inverseRobot.js
```diff
@@ -7,7 +7,30 @@
...
+
+  for (const key of keys) {
```
Viktor-Kost, line 15: We can use robot object directly at this loop 
```suggestion
  for (const key in robot) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,30 @@
...
+    if (count > 1) {
+      return null;
+    }
+
+    count = 0;
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    result[keys[i]] = values[i];
+  }
```
Viktor-Kost, line 31: we need only one condition inside the loop to check current key
```
if(result[robot[key]) {
  return null
}
```

and if this condition is falsy we just update our object
`result[robot[key]] = key;`

Viktor-Kost (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
Good👍

PR #1942, opened 2023-02-13
Solution

ksavelyeva:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robot1 = {};
```
Andr1yK, line 10: robot2,3,4,5.... robotN... 

how to understand what is stored in each of these variables?

Andr1yK (changes requested):

Andr1yK (approved):

PR #1945, opened 2023-02-14
inverseRobot done

Stepan-TS:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const lengthOfRobotObj = Object.keys(robot).length;
+  const clone = {};
+  let lengthOfCloneObj = 0;
```
Andr1yK, line 12: why do u need it as `let`?

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+
+  for (const value in robot) {
+    clone[robot[value]] = value;
+  }
+
+  lengthOfCloneObj = Object.keys(clone).length;
```
Andr1yK, line 18:
```suggestion

  for (const value in robot) {
    clone[robot[value]] = value;
  }

  const lengthOfCloneObj = Object.keys(clone).length;
```

Andr1yK (approved):

PR #1946, opened 2023-02-14
solution

arkushenko-anna:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inverseObj = {};
+
+  for (const key in robot) {
+    if (inverseObj.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    inverseObj[robot[key]] = key;
```

illiarerun (approved):

PR #1950, opened 2023-02-14
solution

chernenko2:


src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
 
 function inverseRobot(robot) {
   // write code here
+
+  const inverse = {};
```
illiarerun, line 12: inverse is a verb, `inversedRobot` is a noun, please use nouns for constants

illiarerun (approved):

PR #1953, opened 2023-02-14
Solution done

yaroslav-vasilyev:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
```
Andr1yK, line 10: not the best name🤔

[NAMING] - use proper names for variables - if you want to name variable `result`(`arr`, `obj`), pick a better one :)

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const [value, key] of Object.entries(robot)) {
+    if (key in newObj) {
+      return null;
+    }
```
Andr1yK, line 15: why do u need to use Object.entries here?

Avoid increasing complexity))

Andr1yK (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (fixedRobot.hasOwnProperty(key)
```
Andr1yK, line 13: as for me... redundant
```suggestion
    if (
```

Andr1yK (changes requested):

Andr1yK (approved):

PR #1956, opened 2023-02-15
Solution

kostiantin-hudzenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const key in robot) {
+    if (inversed.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    inversed[robot[key]] = key;
```
IvanFesenko, line 17: you repeat `robot[key]` 2 times here, better create variable with this value
```suggestion
    const value = robot[key];
    if (inversed.hasOwnProperty(value)) {
      return null;
    }

    inversed[value] = key;
```

IvanFesenko (approved):
Good job. Check my comment

PR #1971, opened 2023-02-20
inverseRobot was created

Pavlo-Petrashevskyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+    if (TRUE_KEYS[FIRST_KEY] === TRUE_KEYS[i]) {
+      return null;
+    }
+
+    FIRST_KEY++;
+  }
+
+  for (let y = 0; y < TRUE_KEYS.length; y++) {
+    OBJ[TRUE_KEYS[y]] = TRUE_VALUES[y];
+  }
```
ViktorHavryliuk, line 26: You can do this much simpler and with less code using for...of loop:
```
for(const key of robot) {
  <your code>
}
```

ViktorHavryliuk (changes requested):

src/inverseRobot.js
Pavlo-Petrashevskyi: I`ve tried to rewrite it using for...of loop, but faced a problem that robot is not iterable. I solved this but i am not  so sure that it is an appropriate variant.

vlad-demchuk: Sorry, I guess, Victor meant `for...in` loop. it is used `for...in` loop for objects.

vlad-demchuk (changes requested):
You could try with `for...in` loop🙂

src/inverseRobot.js
Pavlo-Petrashevskyi: Thanks), I have solved it with both for...of loop (first solution) and for...in loop (second solution)

loralevitska (approved):
good job!

PR #1972, opened 2023-02-20
Solution for Inverse Robot

ctk12:


qoa11a (changes requested):
there's no need for 2 loops

it can be done the next way:

1. creating newObj variable, just how you did
2. iteration over keys in robot using for in loop
3. checking if there's newObj[robot[key]]. If it's there than return null
4. otherwise, newObj[robot[key]] = key
5. return newObj

qoa11a (approved):

PR #1975, opened 2023-02-21
Solution

VovaKlimyuk:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+  const resualt = {};
```
vkryvytskyy, line 12:
```suggestion
  const result = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (let i = 0; i < keys.length; i++) {
```
vkryvytskyy, line 14:
```suggestion
  for (let key in robot) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+    resualt[keys[i]] = values[i];
```
vkryvytskyy, line 15: check here if result already have key that you're trying to add - return null, else - add reversed key

vkryvytskyy (approved):

PR #1977, opened 2023-02-21
js inverse robot

valeriy-mikheyev:


slavanosatskyi (changes requested):
Hey, your approach is right, you just need to change the way you check if key is duplicated. Try to use "in" operator, or hasOwnProperty

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty([robot[key]])) {
+      return null;
```
etojeDenys, line 15:
```suggestion
    if (newRobot[robot[key]])) {
      return null;
```

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
...
+    } else {
+      newRobot[robot[key]] = key;
+    }
```
etojeDenys, line 18:
```suggestion
    } 
      newRobot[robot[key]] = key;
```

etojeDenys (changes requested):

vlad-demchuk (approved):

PR #1979, opened 2023-02-22
reverse robot

DmytroHorshunov:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
-  // write code here
+  const NewRobot = {};
+
+  for (const key in robot) {
+    NewRobot[robot[key]] = key;
+  }
+
+  if (Object.keys(NewRobot).length !== Object.values(robot).length) {
+    return null;
+  }
```
ViktorHavryliuk, line 18: You can make it bit simpler
```suggestion
  for (const key in robot) {
    if (robot[key] in newRobot) {
      return null
    }
  
    newRobot[robot[key]] = key;
  }
```

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const NewRobot = {};
```
ViktorHavryliuk, line 10: variables should be in camelCase
```suggestion
  const newRobot = {};
```

ViktorHavryliuk (changes requested):

src/inverseRobot.js
DmytroHorshunov: ok, thanks

Viktor-Kost (approved):
Nice!

PR #1981, opened 2023-02-23
Solution

StepuraSerhii:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  let result = {};
+
+  for (const key in robot) {
+    if (robot[key] in result) {
+      result = null;
+      break;
```
sTorba24, line 15:
```suggestion
      return null
```

sTorba24 (changes requested):
you should return null, not set result to null

Viktor-Kost (approved):
Nice!

PR #1983, opened 2023-02-23
Solution

1veee:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (invertedRobot[value] !== undefined) {
+      return null;
+    }
+    invertedRobot[value] = key;
```
ViktorHavryliuk, line 16: Small suggestion on how to make it better. Also please pay attention to formatting
```suggestion
    if (value in invertedRobot) {
      return null;
    }
    
    invertedRobot[value] = key;
```

ViktorHavryliuk (approved):
Good work

PR #1985, opened 2023-02-25
first version

Vladimir-Zadorozhnyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (result[value] !== undefined) {
```
TarasHoliuk, line 15: A small suggestion to make it safer (just FYI):
```suggestion
    if (result.hasOwnProperty(value)) {
```

TarasHoliuk (approved):

src/inverseRobot.js
Vladimir-Zadorozhnyi: Task was approve, but in app has such view. 
изображение

OleziO: Not resolved. If you have any questions - feel free to ask them on `fe_chat`.

OleziO (changes requested):

Anastasiia-Svintsova (approved):

PR #1991, opened 2023-02-27
complete task

VladGraund:


src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
...
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = Object.entries(robot).map(([a, b]) => [b, a]);
+
+  for (let i = 0; i < newRobot.length - 1; i++) {
+    if (newRobot[0][0] === newRobot[i + 1][0]) {
```
slavanosatskyi, line 13: You should not rely on order of keys in objects. They are not always sorted as expected. You can check my example from a screenshot. It is better to check whether such key is already present in an array 
Screenshot 2023-02-27 at 17 48 36

slavanosatskyi (changes requested):

src/inverseRobot.js
VladGraund: Should I rebuild a "for" block?

src/inverseRobot.js
```diff
@@ -4,10 +4,23 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.keys(robot);
+
+  for (let i = 0; i < keys.length; i++) {
+    for (let j = i + 1; j < keys.length; j++) {
+      if (robot[keys[i]] === robot[keys[j]]) {
+        return null;
+      }
+    }
```
DarkMistyRoom, line 17: this loop has no sense, cause keys in objects never repeat

DarkMistyRoom (changes requested):
you should create loop through robot keys and check if `newRobot[robot[key]]` is exist before adding. If you have questions, write in the chat

src/inverseRobot.js
```diff
@@ -4,10 +4,23 @@
...
-  // write code here
+  const values = Object.values(robot);
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = i + 1; j < values.length; j++) {
+      if (values[i] === values[j]) {
+        return null;
+      }
+    }
+  }
```
IvanFesenko, line 18: Not fixed previous comment.

Use for in loop for robot, and add to new object

IvanFesenko (changes requested):

src/inverseRobot.js
```diff
@@ -1,13 +1,23 @@
...
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    if (reversedRobot[robot[key]] !== undefined) {
```
OlEzhA0, line 14:
```suggestion
    if (reversedRobot[robot[key]]) {
```

OlEzhA0 (approved):
Gj!

PR #1992, opened 2023-02-27
initial commit

NikolayGit1:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result[robot[key]] === undefined) {
+      result[robot[key]] = key;
+    } else {
+      return null;
```
etojeDenys, line 16:
```suggestion
    if (result[robot[key]]) {
      return null;
    } 
      result[robot[key]] = key;
```

etojeDenys (changes requested):
almost done

src/inverseRobot.js
AlisaVlasova: Hi, I see you haven't fixed previous comments, maybe you did not push your changes?
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster because your answers in GitHub will be visible to the mentor only after you re-request review 🙂

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #1995, opened 2023-02-28
Solution

NVlasenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+    for (let j = 0; j < i; j++) {
+      if (values[j] === values[i]) {
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    inverse[robot[key]] = key;
+  }
```
tetiana-ivanchenko, line 23:
```suggestion
  for (const key in robot) {
    if (robot[key] in inverse) {
      return null;
    }
    
    inverse[robot[key]] = key;
  }
```

tetiana-ivanchenko (changes requested):

src/inverseRobot.js
NVlasenko: @tetiana-ivanchenko  Fixes

sTorba24 (approved):
GJ 🚀

PR #1998, opened 2023-03-01
Robo solution

92BS:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const roboNew = {};
```
```suggestion
  const reversedRobot = {};
```
Try to provide more descriptive names for your variables

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const key in robot) {
+    if (roboNew[robot[key]]) {
```
polosanya, line 13: image

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+      return null;
+    } else {
+      roboNew[robot[key]] = key;
```
polosanya, line 16: image

polosanya (changes requested):
Almost done!

92BS:
Used method hasOwnProperty() 
removed else from this solution

qoa11a (approved):

PR #2000, opened 2023-03-01
solution

vlad-fil:


src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
 
 function inverseRobot(robot) {
   // write code here
```
OleziO, line 10: Remove comment

src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
...
+  const newObj = {};
+  const keys = Object.values(robot);
```
OleziO, line 12: Rename `keys` to `values`

src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
...
+
+  for (let i = 0; i < keys.length; i++) {
```
OleziO, line 14: Need to make `i < keys.length - 1` cause if you have `j = i + 1` at the last element, index `j` - will be out of array indexes.

OleziO (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+    for (let j = i; j < values.length - 1; j++) {
+      if (values[i] === values[j + 1]) {
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    newObj[robot[key]] = key;
+  }
```
etojeDenys, line 23:
```suggestion

  for (const key in robot) {
    if (newObj[robot[key]]) {
     return null;
    }
    newObj[robot[key]] = key;
  }
```

etojeDenys (changes requested):
also, you can try to use in if `hasOwnProperty` instead of `newObj[robot[key]]`

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newObj = {};
+
+  for (const key in robot) {
+    if (newObj.hasOwnProperty(robot[key])) {
```
OlEzhA0, line 13: nb: 
```suggestion
    if (newObj[robot[key]]) {
```

OlEzhA0 (approved):

PR #2005, opened 2023-03-04
task complete

HYPER2307:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
```
OleziO, line 7: Remove all comments from the code.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (inverse[value] !== undefined) {
```
OleziO, line 13: Replace with `inverse[value]`, it will be the same as `inverse[value] !== undefined`

OleziO (changes requested):

nazarmatsevych (approved):

PR #2008, opened 2023-03-05
Inverse is ready

vanya-kalyenichenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const correctVerion = {};
```
Volodymyr-Melnyk-Fron10der, line 10:
```suggestion
  const reversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  for (const key in robot) {
+    const i = robot[key];
```
Volodymyr-Melnyk-Fron10der, line 13:
```suggestion
    const value = robot[key];
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great job! Just pay more attention for names please

IvanFesenko (approved):
Well done

PR #2010, opened 2023-03-05
create inverse robot

allagutsul27:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+    for (let j = i + 1; j < values.length; j++) {
+      if (values[i] === values[j]) {
+        return null;
+      } else {
+        for (const key in robot) {
+          inverse[robot[key]] = key;
+        }
+      }
+    }
+  }
```
etojeDenys, line 24: let's rewrite it just using `for of` (`Object.keys(robot)`) and checking if the property already exists in `inverse` with `.hasOwnProperty()`

etojeDenys (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const inverse = {};
+
+  const keys = Object.keys(robot);
+
+  for (const key of keys) {
+    if (inverse.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      inverse[robot[key]] = key;
+    }
```
IvanFesenko, line 19: else statement is redudant
```suggestion
    if (inverse.hasOwnProperty(robot[key])) {
      return null;
    }
    
    inverse[robot[key]] = key;

```

IvanFesenko (approved):
Well done. Check my comment to avoid same mistakes in the future

src/inverseRobot.js
allagutsul27: Yes. I agree. Thank you for comment!

PR #2014, opened 2023-03-06
Develop

Vaskovics:


src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+    const keyArr = Object.keys(robot);
+
+    delete robot[key];
+
+    if (keyArr.includes(newKey)) {
+      return null;
+    }
+
+    robot[newKey] = newValue;
```
ivbutsykin, line 25: No need to mutate initial value. Better to create new object and add values here. In such way your code will be more descriptive and simple: 
```
const result = {};
let newKey;
let newValue;

for (const key in robot) {
  ...
  result[newKey] = newValue;
}

return result;
```

ivbutsykin (changes requested):
Try to rewrite code without `robot` mutation

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  let newKey;
+  let newValue;
+  const result = {};
+
+  for (const key in robot) {
+    newKey = robot[key];
+    newValue = key;
```
IvanFesenko, line 16: You can create it inside loop
```suggestion
  const result = {};

  for (const key in robot) {
    const newKey = robot[key];
    const newValue = key;
```

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+
+    const keyArr = Object.keys(result);
+
+    if (keyArr.includes(newKey)) {
+      return null;
+    }
```
IvanFesenko, line 22: To complicated
Just do
```
if (result[newKey]) {
  return null;
}
```

IvanFesenko (changes requested):
Check my comments

src/inverseRobot.js
Vaskovics: i was thinking that is better to create  variable or const outside the loop

SanyaBratashchuk (approved):

PR #2016, opened 2023-03-06
my solution

hyrych-max:


src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
 
 function inverseRobot(robot) {
   // write code here
```
OleziO, line 10: Remove comment.

src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
...
+  const correctRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (correctRobot[value] !== undefined) {
```
OleziO, line 16: Replace with `correctRobot[value]`, it will be the same as `correctRobot[value] !== undefined`

OleziO (approved):

PR #2020, opened 2023-03-07
Develop

domkab:


src/inverseRobot.js
```diff
@@ -1,13 +1,33 @@
...
-/*
- * @param {object} robot
- *
- * @return {object}
-*/
+const robert = {
+  Robert: 'name',
+  123: 'chipVer',
+  113: 'chipVerPro',
+};
```
polosanya, line 7: You don't need to create it manually. It's just an example

src/inverseRobot.js
```diff
@@ -1,13 +1,33 @@
...
+
+inverseRobot(robert);
 
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  if (robot.hasOwnProperty('name') && robot.name !== 'name') {
+    inversedRobot[robot.name] = 'name';
+  }
```
polosanya, line 16: Why do you need to check the name separately? 
Can't you check it with all other properties?

src/inverseRobot.js
```diff
@@ -1,13 +1,33 @@
...
+
+  for (const objectKey in robot) {
+    const objectValue = robot[objectKey];
```
polosanya, line 19:
```suggestion
  for (const key in robot) {
    const value = robot[key];
```
No need to add "object" before name

src/inverseRobot.js
```diff
@@ -1,13 +1,33 @@
...
+
+    if (objectKey !== 'name') {
+      if (Object.keys(inversedRobot).includes(objectValue)) {
+        return null;
+      }
+    }
```
polosanya, line 25:
```suggestion
    if (inversedRobot.hasOwnProperty(value)) {
      return null
    }
```
image

polosanya (changes requested):
Let's improve your code a bit)

src/inverseRobot.js
```diff
@@ -1,13 +1,25 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (robot.hasOwnProperty('name') && robot.name !== 'name') {
+      inversedRobot[robot.name] = 'name';
+    }
```
anstsot, line 11: why the `name` is hardcoded here? `robot` can have value with any naming. so just check `key` and `value` from line 6 and 7
```suggestion
    if (robot.hasOwnProperty(value)) {
      inversedRobot[value] = key;
    }
```

anstsot (changes requested):
please remove the hardcoding and let the function work with any `key`/`value` pairs.

src/inverseRobot.js
domkab: because here I tried to make property name always come first, for some reason it didnt go as first property and i failed all tests

src/inverseRobot.js
```diff
@@ -1,13 +1,31 @@
...
+
+    if (robot.hasOwnProperty(value)) {
+      inversedRobot[value] = key;
+    }
+
+    if (inversedRobot.hasOwnProperty(value)) {
+      return null;
+    }
+
+    inversedRobot[value] = key;
```
polosanya, line 25:
```suggestion
    if (inversedRobot.hasOwnProperty(value)) {
      return null;
    }

    inversedRobot[value] = key;
```
Do you really need a first check?

polosanya (changes requested):
Almost perfect!

anstsot (approved):
Great job, approved!

PR #2023, opened 2023-03-09
Solution

pindel-kateryna:


src/inverseRobot.js
```diff
@@ -8,6 +8,31 @@
...
+
+  if (repeat === false) {
+    for (const key in robot) {
+      inversedRobot[robot[key]] = key;
+    }
+
+    return (inversedRobot);
+  } else {
+    return (null);
+  }
```
IvanFesenko, line 35: It's to complicated. You should use for of loop for robot, check if inversedRobot already has value as key - if so - return null other way - add new value

IvanFesenko (changes requested):
Let's refactor. If you have questions - ask them in fe_chat

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (inversedRobot.hasOwnProperty(robot[key])) {
+      return null;
```
sTorba24, line 15:
```suggestion
      return
```

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
...
+    } else {
+      inversedRobot[robot[key]] = key;
+    }
```
sTorba24, line 18:
```suggestion
    } 
    
    inversedRobot[robot[key]] = key;
   
```

sTorba24: not needed to use else here :)

sTorba24 (approved):
Good job!

PR #2025, opened 2023-03-09
function to reverse object value

DenysKrysanov:


prokivskaaa (approved):
Great! Everything is alright!

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
```
```suggestion
  const reversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (value in reverseRobot) {
+      return null;
+    }
+    reverseRobot[value] = key;
```
```suggestion
    }

    reverseRobot[value] = key;
```

jstmpelowycz (approved):

PR #2033, opened 2023-03-13
Solution

PashaMalyshkin:


Zad0rnik (approved):
GJ PASHA ХТЗ РУЛИТ

PashaMalyshkin:
> GJ PASHA ХТЗ РУЛИТ

HAHAHAH

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+  const robotKey = Object.values(robot);
+  const robotValue = Object.keys(robot);
```
jstmpelowycz, line 12: even if you DO need to use it, let's give it proper names
```suggestion
  const robotKeys = Object.values(robot);
  const robotValues = Object.keys(robot);
```

jstmpelowycz: however, it'd be better to use simple for-in-loop — consistent and readable

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  if (new Set(robotKey).size !== robotKey.length) {
+    return null;
+  }
```
jstmpelowycz, line 16: interesting approach

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #2036, opened 2023-03-13
Solution

Zad0rnik:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+    const value = robot[key];
+
+    if (repairedRobot.hasOwnProperty(value)) {
+      return null;
+    }
+
+    repairedRobot[value] = key;
+  }
+
+  return repairedRobot;
```
PavloReutskiy, line 22:
```suggestion
  const repairedRobot = {};

  for (const key in robot) {
    if (repairedRobot.hasOwnProperty(robot[key])) {
      return null;
    }

    repairedRobot[robot[key]] = key;
  }

  return repairedRobot;
```

you don't have to use this variable, but I'm not sure that it's important

PavloReutskiy (approved):
🚀

jstmpelowycz (approved):

PR #2045, opened 2023-03-13
create inverseRobot

VenomMB:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const arrOfKeys = Object.values(robot).reverse();
+  const arrOfValue = Object.keys(robot).reverse();
```
Olekk17, line 11: You can use Object.entries instead of using two separate functions and two separate variables

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const inverseObjectRobot = {};
+
+  const lenOfArrays = arrOfKeys.length;
```
Olekk17, line 14: I am not sure, but you don't need blank lines between variables declaration

Olekk17 (approved):
Good job

src/inverseRobot.js
VinnieJ-2k20: Yep, better to use entries to keep key and value tied together.

src/inverseRobot.js
VinnieJ-2k20: You use the value just once, no need for extra variable.

VinnieJ-2k20 (changes requested):
Your solution works, but can be improved.

VinnieJ-2k20 (approved):

PR #2046, opened 2023-03-13
Solution

KosolapovAnatolii:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const k in robot) {
```
Anastasia2403, line 12: It`s not clear what this k-variable means, change it to a clear one (checklist point 3).

Anastasia2403 (changes requested):
Great solution 👍 But pay attention to my comments.

VinnieJ-2k20 (approved):
Good job 👍

PR #2047, opened 2023-03-13
done

victor-buldenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotValues.length; i++) {
+    if (robotValues.indexOf(robotValues[i])
+      !== robotValues.lastIndexOf(robotValues[i])) {
+      return null;
+    }
+  }
```
Volodymyr-Melnyk-Fron10der, line 17: but i don't get the purpose of this piece of code actually) remove it

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+
+  const robotEntries = Object.entries(robot);
+
+  robotEntries.map(el => el.reverse());
```
Volodymyr-Melnyk-Fron10der, line 21: amazing solution! just do it in right way. map is not mutating the array, it returns new one
```suggestion
  const reversedEntries = Object.entries(robot).map(el => el.reverse());
```

Volodymyr-Melnyk-Fron10der (changes requested):
Keep it up!

src/inverseRobot.js
victor-buldenko: i don`t know what i must to do... have any ideas? In my mind it`s most obvious solution.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+
+  for (let i = 0; i < robotValues.length; i++) {
+    if (i !== robotValues.lastIndexOf(robotValues[i])) {
+      return null;
+    }
```
SerhiiKirik, line 15:
```suggestion
  for (let index = 0; index < robotValues.length; index++) {
    const lastIndexValue = robotValues.lastIndexOf(robotValues[index]);
    
    if (index !== lastIndexValue)) {
      return null;
    }
```

SerhiiKirik (approved):
Great 🔥

PR #2048, opened 2023-03-13
function inverseRobot was created

Vasyl-Zhyliakov:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 19: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
-  // write code here
+  let inverse = {};
+
+  for (const key in robot) {
+    if (inverse[robot[key]]) {
+      inverse = null;
+      break;
+    }
+
+    inverse[robot[key]] = key;
```
SerhiiKirik, line 18:
```suggestion
  for (const key in robot) {
    const robotKey = robot[key];
    
    if (inverse[robotKey]) {
      inverse = null;
      break;
    }

    inverse[robotKey] = key;
```

SerhiiKirik (approved):
Well done 🚀

PR #2049, opened 2023-03-13
Solution

Stambuliak:


Vlad-Lischenovskiy (approved):
Interesting method, Good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.keys(robot);
+  const values = Object.values(robot);
+  const withoutRepeat = new Set(values);
```
jstmpelowycz, line 12:
```suggestion
  const uniqueValues = new Set(values);
```

jstmpelowycz (approved):
honestly, complicated approach, for-in would fit better

PR #2050, opened 2023-03-13
made a solution

Olekk17:


danylo-dobrovolskyi (approved):
Good job!

src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
...
+  const newRobot = {};
+  const robotEntries = Object.entries(robot);
+
+  for (let i = 0; i < robotEntries.length; i++) {
+    for (let j = i; j < robotEntries.length; j++) {
+      if (robotEntries[i][1] === robotEntries[j][1] && i !== j) {
+        return null;
+      }
+    }
+  }
```
VinnieJ-2k20, line 20: Don't write nested loops unless completely necessary.
Think on how to perform the same check in yout for ... of loop.

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):
Good job 👍 
Could use `for ... in` loop and avoid extra array.

PR #2052, opened 2023-03-13
Solution

Anna-Kuzmenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const correctRobot = {};
+
+  for (const key in robot) {
+    if (correctRobot[robot[key]]) {
```
Rage1502, line 13: You need to use "hasOwnProperty". This is the second item on the checklist.

Rage1502 (changes requested):

src/inverseRobot.js
Anna-Kuzmenko: fixed

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const correctRobot = {};
+
+  for (const key in robot) {
+    if (correctRobot.hasOwnProperty([robot[key]])) {
+      return null;
+    }
+    correctRobot[robot[key]] = key;
```
jstmpelowycz, line 16: the checklist, p.1

jstmpelowycz (changes requested):

src/inverseRobot.js
Anna-Kuzmenko: I added a variable, but I don't understand is it right or not (

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const correctRobot = {};
+
+  for (const key in robot) {
+    const property = robot[key];
```
jstmpelowycz, line 13: IMHO, `property` is something more abstract, combining both key and value
```suggestion
    const value = robot[key];
```

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+
+    if (correctRobot.hasOwnProperty([property])) {
+      return null;
+    }
+    correctRobot[property] = key;
```
```suggestion
    }

    correctRobot[property] = key;
```

jstmpelowycz (approved):
good job!

PR #2055, opened 2023-03-14
reversed robots keys

Vlad-Lischenovskiy:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (reverseRobot[value] !== undefined) {
```
VladislavaT, line 15: I think we have this in checklist
it`s better to use hasOwnProperty insted of  !== undefined

VladislavaT (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
```
```suggestion
  const reversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (reverseRobot.hasOwnProperty(value)) {
+      return null;
+    }
+    reverseRobot[value] = key;
```
```suggestion
    }

    reverseRobot[value] = key;
```

jstmpelowycz (approved):

PR #2061, opened 2023-03-14
Solution

Kavon-Skela:


M1k1ta (approved):
Good job. This task can solve writing less code. I think it's enough when it works and you follow the rules from checklist.

src/inverseRobot.js
```diff
@@ -7,7 +7,29 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const uniqueValues = new Set(Object
+    .values(robot))
+    .size;
+  const allValues = Object
+    .values(robot)
+    .length;
```
jstmpelowycz, line 15:
```
  
  if (values.length !== uniqueValues.size) {
  ...
```

src/inverseRobot.js
```diff
@@ -7,7 +7,29 @@
...
+
+  if (uniqueValues !== allValues) {
+    return null;
+  }
+
+  const invertedRobot = { ...robot };
```
jstmpelowycz, line 21: this is extra, you don't have to do it, it gives complexity

jstmpelowycz (changes requested):
let's use another, simpler approach, write values to your new object, do not delete/overwrite

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+
+  const uniqueValues = new Set(values);
```
jstmpelowycz, line 12:
```suggestion
  const values = Object.values(robot);
  const uniqueValues = new Set(values);
```

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+  if (uniqueValues.size !== values.length) {
+    return null;
+  }
+
+  const invertedRobot = { };
```
```suggestion
  const invertedRobot = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (!Object.prototype.hasOwnProperty.call(invertedRobot, key)) {
+      invertedRobot[value] = key;
+    }
```
jstmpelowycz, line 25: you don't have to validate this condition, it's handled by the condition above
btw, `hasOwnProperty` call could've been simplified, no need to access via prototype
```
const object = {...};

if (object.hasOwnProperty(...)) {
...
```

jstmpelowycz (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+  if (uniqueValues.size !== values.length) {
+    return null;
+  }
+
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    invertedRobot[value] = key;
```
jstmpelowycz, line 22: in this case, you could've avoided using variable
why? you use this value once, so there's no actual purpose

jstmpelowycz (approved):

PR #2066, opened 2023-03-14
solution

Nazarii98:


makandriival (approved):
Good Job 👍 . I think my solution a bit more elegant. Lmk what do you think

`
  const newRobot = {};

  for (const elem in robot) {
    if (robot[elem] in newRobot) {
      return null;
    } else {
      newRobot[robot[elem]] = elem;
    }
  }

  return newRobot;
}
`

VinnieJ-2k20 (changes requested):
Solution works, but can be improved.

src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
 
 function inverseRobot(robot) {
   // write code here
+  const trueRobot = {};
+
+  for (const key in robot) {
+    const paramName = robot[key];
```
jstmpelowycz, line 14: it refer neither to param, nor to name, it's just a value
```suggestion
    const value = robot[key];
```

src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
...
+
+    if (robot[key] in trueRobot) {
```
jstmpelowycz, line 16: don't forget to use your variable here

jstmpelowycz (changes requested):

src/inverseRobot.js
jstmpelowycz: 😢

jstmpelowycz (changes requested):

src/inverseRobot.js
Nazarii98: I'm sorry, don't understand what I should do here, I changed the name for my variable everywhere, how i can use this variable here?

Nazarii98: Understood

jstmpelowycz (approved):

PR #2068, opened 2023-03-14
task is work

M1k1ta:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+
+  for (const key in robot) {
+    const rightKey = robot[key];
+    const rightValue = key;
+
+    if (clearSetting.hasOwnProperty(rightKey)) {
+      return null;
+    }
+
+    clearSetting[rightKey] = `${rightValue}`;
```
Evgenb000, line 20:
```suggestion
  for (const key in robot) {
    const rightKey = robot[key];

    if (clearSetting.hasOwnProperty(rightKey)) {
      return null;
    }

    clearSetting[rightKey] = key;
```

Evgenb000: I think that`s looks better

Evgenb000 (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const clearSetting = {};
+
+  for (const key in robot) {
+    const rightKey = robot[key];
+    const rightValue = key;
```
VinnieJ-2k20, line 14: Yep, just use avaikable `key` variable, no need to introduce new one.

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const clearSetting = {};
```
VinnieJ-2k20, line 10: Not the best variable name.

VinnieJ-2k20 (approved):
Good job 👍 
Read the comments.

PR #2071, opened 2023-03-14
Robot mirrored

roshirin:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const mirrorRobot = {};
```
Goto-Jesus, line 10: I think it is not necessary to create a variable. 

You can change the object (robot) because if you use this object in the function it is another object which exists only in this function)))

But it is not a mistake)) Good job :D

Goto-Jesus (approved):

src/inverseRobot.js
roshirin: Manipulations with robot parameter inside the function will mutate the initial object. And we have a request to return a new object in our task.

jstmpelowycz: @roshirin is totally right

jstmpelowycz (approved):
great job 🔥

PR #2074, opened 2023-03-15
Solution

makandriival:


jstmpelowycz (changes requested):
the checklist, p.1, 4

src/inverseRobot.js
```diff
@@ -8,6 +8,16 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const elem in robot) {
+    if (robot[elem] in newRobot) {
+      return null;
+    }
+    newRobot[robot[elem]] = elem;
```
jstmpelowycz, line 17:
```suggestion
    if (robot[elem] in newRobot) {
      return null;
    }

    newRobot[robot[elem]] = elem;
```

jstmpelowycz: also, let's use a variable for the repeating `robot[elem]`

src/inverseRobot.js
```diff
@@ -8,6 +8,16 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+
+  for (const elem in robot) {
```
jstmpelowycz, line 13: `elem` :harold:
image

jstmpelowycz (changes requested):

src/inverseRobot.js
makandriival: hehe

jstmpelowycz (changes requested):
> also, let's use a variable for the repeating robot[elem]

what about this one?

src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
...
+    const element = robot[elem];
+
+    if (element in newRobot) {
+      return null;
+    };
+    newRobot[element] = elem;
```
jstmpelowycz, line 19:
```suggestion
    const value = robot[key];

    if (value in newRobot) {
      return null;
    };

    newRobot[value] = key;
```

jstmpelowycz (approved):

PR #2075, opened 2023-03-15
function works

BondarenkoDmitriy:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (reversRobot[value]) {
```
StasLukianchuk, line 15: p.2 - use correct check for property presence in object. Some properties can be present but still contain falsy values.

StasLukianchuk (changes requested):
GJ

src/inverseRobot.js
BondarenkoDmitriy: I don't understand what check you mean.

BondarenkoDmitriy: okey, I think I mean it about .hasOwn

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversRobot = {};
```
```suggestion
  const reversedRobot = {};
```

jstmpelowycz (approved):

PR #2084, opened 2023-03-15
version1

AlexRog01:


src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValuesMas = Object.values(robot);
+  const originalRobotValues = [];
+
+  for (const curr of robotValuesMas) {
+    if (originalRobotValues.includes(curr)) {
```
jstmpelowycz, line 14: let's make it simple
you don't actually have to create a separate array to check what's inside your object
for example, you can use `hasOwnProperty` method

jstmpelowycz (changes requested):
good job, everything works! see the comment 👇

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const changedObj = {};
+
+  for (const key in robot) {
+    if (changedObj.hasOwnProperty(robot[key])) {
+      return null;
+    }
+    changedObj[robot[key]] = key;
```
jstmpelowycz, line 16: formatting + the checklist, p.1
```suggestion
    if (changedObj.hasOwnProperty(robot[key])) {
      return null;
    }

    changedObj[robot[key]] = key;
```

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const changedObj = {};
```
```suggestion
  const newRobot = {};
```

jstmpelowycz (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (inversedRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    inversedRobot[robot[key]] = key;
```
jstmpelowycz, line 17: the checklist, p.1

jstmpelowycz (changes requested):

src/inverseRobot.js
AlexRog01: I don't know what else to change here.

liliia-livandovskaya (approved):

PR #2087, opened 2023-03-16
Solution

DanielSorokowski:


src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  let array = [];
```
solaryasha, line 10: Avoid reassigning the variable values

solaryasha: Avoid using the data structure name in the variable name. It's kind of strange :)

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+  // Change object into one array
+  Object.entries(robot).forEach(par => {
+    array = array.concat(par);
```
solaryasha, line 14: You could just use `push` method. It's more efficient

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+  if (checkDuplicates) {
+    return null;
+  }
+
+  // make arrays from objects keys and swap them;
+  const fixRobot = Object.entries(robot).map(([key, value]) => [value, key]);
+
+  // change arrays into object and return
+  return Object.fromEntries(fixRobot);
```
solaryasha, line 28: It's a working solution. But let's think about how to solve it easier.

1. you could create the empty object
2. iterate through the existing object using `for in` loop
3. check the value stored in the object by the current key. 
4. If you have a such key in the inversed object, then return null
5. If not, just set in the inversed key

One loop solution without any additional arrays.

solaryasha (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedObject = {};
```
```suggestion
  const invertedRobot = {};
```
Avoid using data structures in the naming of the variables

solaryasha (changes requested):

solaryasha (changes requested):
.

solaryasha (approved):

PR #2091, opened 2023-03-17
develop

NataliiaRodoman:


src/inverseRobot.js
```diff
@@ -7,7 +8,27 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  for (const key in robot) {
+    const property = robot[key];
+    let count = 0;
+
+    for (const key2 in robot) {
```
mykhalenych, line 15: circle inside another circle it isn't good practice, try to init object and in first circle check ```
yourVariable.hasOwnProperty(robot[key]) {
      return null;
}
if not yourVariable[robot[key]] = key;
```

**Review by mykhalenych — CHANGES_REQUESTED** [review]

File: src/inverseRobot.js [hunk 2]
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-param-reassign */
```
OleziO, line 1: Remove all comments.

OleziO (approved):

PR #2092, opened 2023-03-18
Solution

skinnyjazz:


generatedman (approved):
good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverted = {};
+
+  for (const prop in robot) {
+    if (inverted[robot[prop]] !== undefined) {
```
naziamoff, line 13: this should work the same way if we just convert it to boolean value this way. It seems to me you need just to check whether this property is not empty, not to check whether it equals undefined
```suggestion
    if (inverted[robot[prop]]) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+      return null;
```
naziamoff, line 14: do you need to return null here? Seems `break` or `continue` might come in handy here

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverted = {};
```
naziamoff, line 10: invertedWhat?

naziamoff (changes requested):
not bad, but let's apply some polishing

src/inverseRobot.js
naziamoff: disregard this comment, i found out it is required in task description

KatyaZaremska (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
silchyn, line 10:
```suggestion
  const inversedRobot = {};
```

silchyn: check checklist rule `#3`

silchyn (changes requested):

cephy77 (approved):
Nice solution!

silchyn (approved):

naziamoff (approved):

PR #2115, opened 2023-03-29
solution

kq333:


michalcywka (changes requested):
The solution is interesting, but it has one problem. If we find that a value is repeated, we return null in this step and then continue the loop. It would be simpler if we create an empty object first, populate it with values and check every loop iteration if the key we are trying to add already exists in that object. If it does, then we can return null and the loop will stop.

src/inverseRobot.js
```diff
@@ -4,10 +4,26 @@
...
+  for (const [key, value] of Object.entries(robot)) {
+    if (arrayItems.includes(value)) {
+      return null;
+    }
+
+    if (result[value]) {
+      return null;
+    }
+
+    arrayItems.push(value);
```
d-wasilewski, line 22: Do it without using any additional array. use one of checklist [2] examples to check if a property exists

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2116, opened 2023-03-29
solution v1

bjeziorek:


natalia-klonowska (approved):

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const robotToFix = robot;
```
michalcywka, line 11: Do we need to do that? :)

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
...
+  const fixedRobot = {};
+
+  for (const part in robotToFix) {
+    if (fixedRobot.hasOwnProperty(robotToFix[part])) {
+      return null;
+    }
+    fixedRobot[robotToFix[part]] = part;
+  }
+
+  return (fixedRobot);
```
michalcywka, line 21: `()` are not necessary :)

michalcywka (changes requested):

d-wasilewski (approved):

PR #2117, opened 2023-03-29
First solution

marshalek37:


SuaBy514 (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    const info = robot[key];
```
michalcywka, line 13: `info` is not really a good variable name in this case

michalcywka (changes requested):

d-wasilewski (approved):

PR #2121, opened 2023-03-29
Solution

sylcym:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const transponsedRobot = {};
```
d-wasilewski, line 10: There is a small typo in the variable name, it should be transposedRobot instead of transponsedRobot

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const [key, value] of Object.entries(robot)) {
+    transponsedRobot[value] = key;
+  }
+
+  if (Object.keys(robot).length !== Object.keys(transponsedRobot).length) {
+    return null;
+  }
```
d-wasilewski, line 18: Instead of checking for duplicate values after the loop, you can check for duplicates inside the loop. If a duplicate is found, you can return null immediately, making the function more efficient

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2124, opened 2023-03-29
Solution

SuaBy514:


src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inverted = {};
+  const roKeys = Object.keys(robot);
+  const roVals = Object.values(robot);
```
d-wasilewski, line 13: be more descriptive, don't be afraid to use full name, `ro` sounds mysterious

d-wasilewski (changes requested):
r

src/inverseRobot.js
SuaBy514: It was originally this way to avoid the linter being mad about lines being long... although in hindsight it was unnecessary

d-wasilewski (approved):

PR #2126, opened 2023-03-30
add solution

PufAnimalsDev:


src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
...
+    const checkValue = robotValues.shift();
+
+    if (robotValues.includes(checkValue)) {
+      return null;
+    }
+  }
+
+  for (const key in robot) {
+    invertedRobot[robot[key]] = key;
+  }
```
d-wasilewski, line 24: Do the check inside of `for (const key in robot)` loop

d-wasilewski (changes requested):

src/inverseRobot.js
```diff
@@ -8,6 +8,20 @@
 
 function inverseRobot(robot) {
   // write code here
+  const robotValues = Object.values(robot);
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    const checkValue = robotValues.shift();
+
+    if (robotValues.includes(checkValue)) {
```
d-wasilewski, line 17: just do `if (robot[key] in invertedRobot) {`

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2127, opened 2023-03-30
solution

NadieinOleh:


src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const correctRobot = {};
+
+  for (const key in robot) {
+    if (correctRobot[robot[key]]) {
```
witflash, line 13: You can declare const:
```
const value = robot[key];
```
And then it will be more readable:
```
if (correctRobot[value]) {
  ...
}

correctRobot[value] = key;
```
But it is just if you like

witflash (approved):
Good work! 👍🏼

PR #2133, opened 2023-03-31
Solution

Ir-ra:


src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
-  // write code here
+  const newRobot = {};
+  const pairs = Object.entries(robot);
+
+  for (const pair of pairs) {
+    const key = pair[0];
+    const val = pair[1];
+
+    newRobot[val] = key;
+  }
```
agniya-i, line 18: try to use `for...in` to make the solution more simple

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+
+  const keysRobotLength = Object.keys(robot).length;
+  const keysNewRobotLength = Object.keys(newRobot).length;
+
```
agniya-i, line 22: interesting solution 🙂 
instead of comparing by length, in the loop, you can also check if the key is already in the object

agniya-i (changes requested):

OleziO (approved):

PR #2135, opened 2023-04-01
solve

ctgdprzyb:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
+
+  for (const value in robot) {
```
d-wasilewski, line 12: this should be named key for better readability. `value` is misleading

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2141, opened 2023-04-04
inverse robot solution

oleh-dehtiarenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (robot[key] in result) {
+      return null;
+    } else {
+      result[robot[key]] = key;
+    }
```
qoa11a, line 17: You don't need to use else if you return something because a return will immediately exit out of a function.

qoa11a (approved):

src/inverseRobot.js
oleh-dehtiarenko: understood, thanks

PR #2143, opened 2023-04-05
Solution

IvanOrinchak:


Volodymyr-Melnyk-Fron10der (approved):
A clear solution👍 However, it's a good task to practice a reduce method as well

PR #2145, opened 2023-04-05
Solution

BiletskyiVolodymyr:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```

polosanya (changes requested):
Great! One small but important moment left)
Let's get used to giving variables descriptive and meaningful names 🙂

polosanya (approved):
Well done!

PR #2146, opened 2023-04-06
done

maxonshev:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 19: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (!result[robot[key]]) {
```
SerhiiKirik, line 13: you use robot[key] 3 times
Move to the variable

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+      result[robot[key]] = key;
+    } else if (result[robot[key]]) {
```
SerhiiKirik, line 15: you can avoid this `else if` bc if you already checked that "you don't have a key" and it's false, it means that you have this key))

SerhiiKirik, line 15:
```suggestion
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
Great 🔥

PR #2150, opened 2023-04-07
solution

Absolution26:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const objKeys = Object.keys(robot);
+  const objValues = Object.values(robot);
+  const trueRobot = {};
+
+  for (let i = 0; i < objKeys.length; i++) {
+    if (objValues.includes(objValues[i], i + 1)
+    || objKeys.includes(objKeys[i], i + 1)) {
```
agniya-i, line 16: Why don't we just check if the result object already has such key?)

```suggestion
    if (trueRobot.hasOwnProperty(objValues[i])) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const objKeys = Object.keys(robot);
+  const objValues = Object.values(robot);
+  const trueRobot = {};
+
+  for (let i = 0; i < objKeys.length; i++) {
```
agniya-i, line 14: you can also use `Object.entries()` to make code more simple https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

agniya-i (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const objEntries = Object.entries(robot);
```
agniya-i, line 10:
```suggestion
  const robotObjectEntries = Object.entries(robot);
```
give variables more descriptive names

agniya-i (approved):
👍

PR #2151, opened 2023-04-07
Solution

AnastassiaH:


YegorKorenkov (changes requested):
your solution is correct, but please use hasOwnProperty method, it will shorten your code

nazarmatsevych (approved):

PR #2152, opened 2023-04-08
Solution1

RatiBorRozGa:


src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newRobot = {};
+  const seenKeys = {};
```
DarkMistyRoom, line 12: why do you need `seenKeys`? You can check if `newRobot `has current property

DarkMistyRoom (changes requested):
Good! Just one suggestion:

Volodymyr-Melnyk-Fron10der (approved):
🔥

PR #2154, opened 2023-04-09
Solution

mariuszgit:


DorotaLeniecDev (changes requested):
Try using hasOwnProperty object method for checking the duplicates.

Radoslaw-Czerniawski (approved):
Very nice implementation :), take a look at the comment about forEach for future reference

PR #2156, opened 2023-04-09
Inverted Robot solution

VKonoplianyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (key in robot) {
```
SerhiiKirik, line 13: also, you can use hasOwnProperty here

SerhiiKirik (approved):
Well done 🚀

PR #2169, opened 2023-04-14
First commit

Khalniepiesova:


src/inverseRobot.js
```diff
@@ -7,7 +7,28 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  const arrKeys = [];
+
+  for (const key in robot) {
+    arrKeys.push(key);
+  }
```
sTorba24, line 16: you can remove code below and do everything in this loop :)
if key already exists => return null;
directly here set result[robot[key]] = key;

at the end just return result;
`arrKeys` can be removed too;

sTorba24 (changes requested):
Well done! I propose you to simplify it

YegorKorenkov (approved):

PR #2170, opened 2023-04-14
Solution

ZupaFly:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newObject = {};
+
+  for (const key in robot) {
+    if (newObject.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      newObject[robot[key]] = key;
+    }
```
Anastasiia-Svintsova, line 17: `else` is redundant here. The function will stop if the first condition will be true
```suggestion
    }
    
    newObject[robot[key]] = key;
```

Anastasiia-Svintsova (approved):
👍

src/inverseRobot.js
ZupaFly: Understand, thank you!

PR #2171, opened 2023-04-14
Solution

TommasoGiacopini:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
+      return null;
+    } else {
+      newRobot[robot[key]] = key;
+    }
```
Anastasiia-Svintsova, line 17: `else` is redundant here. The function will stop anyway if the first condition will be true

Anastasiia-Svintsova (approved):
👍

PR #2173, opened 2023-04-15
solution

Vladyslav-Voronkov:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const values = Object.values(robot);
+  const keys = Object.keys(robot);
+
+  for (let i = 0; i < values.length; i++) {
```
VitaliyBondarenko1982, line 14:
```suggestion

  for (const key in robot) {
```
consider to use for in loop

and make everything in one loop. 
`const value = robot[key];`

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+    if (values.indexOf(values[i]) !== values.lastIndexOf(values[i])) {
+      return null;
+    }
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    newRobot[values[i]] = keys[i];
+  }
```
VitaliyBondarenko1982, line 22:
```suggestion
```
redundant. make everything in for in loop

VitaliyBondarenko1982 (changes requested):

ivanchuk-yurii (approved):

PR #2175, opened 2023-04-17
Solution

Volodymyr-Terzov:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
```
mykhalenych, line 13: js community recommends using .hasOwn() instead .hasOwnProperty
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

mykhalenych (approved):

PR #2193, opened 2023-04-24
Solution

artpetrenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const newRobot = {};
+  const newKeys = Object.keys(robot).reverse();
+  const newValues = Object.values(robot).reverse();
```
SanyaBratashchuk, line 12: Simplify the solution Object.keys and Object.values is redundant here

SanyaBratashchuk (changes requested):

src/inverseRobot.js
artpetrenko: if I could )) tell me how...

anstsot (approved):
Great job, approved!

PR #2196, opened 2023-04-26
Solution

BeeGooDee:


src/inverseRobot.js
```diff
@@ -7,7 +7,27 @@
...
+    if (checkKeys > -1 || checkValues > -1) {
+      return null;
+    }
+  }
+
+  for (let i = 0; i < values.length; i++) {
+    fixedRobot[values[i]] = keys[i];
+  }
+
+  return fixedRobot;
```
DarkMistyRoom, line 30: This solution is too complicated. You can do it much simpler way by using one loop and calling robot's keys and values directly without converting to an arrays

DarkMistyRoom (changes requested):
Simplify your solution

OleziO (approved):

PR #2199, opened 2023-04-26
solution

1v43k:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+
+  const values = [];
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (values.includes(value)) {
+      return null;
+    } else {
+      inverse[value] = key;
+      values.push(value);
+    }
```
anstsot, line 20: no need to introduce `values` variable. 
```suggestion
    if (inverse.hasOwnProperty(value)) {
      return null;
    }
    
    inverse[value] = key;
```

anstsot (changes requested):
t’s almost done! Let’s improve a little

PR #2200, opened 2023-04-27
Solution

VolodymyrPst:


src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const arr = [];
```
polosanya, line 10: It would be great to use more meaningful names

src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
...
+
+  for (const i of Object.entries(robot)) {
```
polosanya, line 12:
```suggestion
  for (const key in robot) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
...
+    arr.push(i.reverse());
+  };
+
+  const res = Object.fromEntries(arr);
+
+  return Object.keys(res).length < Object.values(robot).length ? null : res;
```
polosanya, line 18: It looks too complicated. Consider making it easier.
Just iterate through robot's keys and create a reversed property in a new robot. 
If it's already present - return null, that's all

polosanya (changes requested):
Good job! Let's try to improve your solution)

PR #2201, opened 2023-04-27
Solution

vintwp:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+  const robotCorrect = {};
+
+  for (let i = 0; i < robotValues.length; i++) {
```
YegorKorenkov, line 13: Here you can iterate not over your robotValues  array, but immediately over the robot array. And in a loop check if there is a field in the result object that is equal to robot[key] then return null, if not, then add the desired field with the desired value. This will shorten your code and allow you not to use the loop again

YegorKorenkov (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotCorrect = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (robotCorrect[value]) {
```
Volodymyr-Melnyk-Fron10der, line 13: Instead of checking if `robotCorrect[value]` exists, use `robotCorrect.hasOwnProperty(value)` to avoid issues with inherited properties.

Volodymyr-Melnyk-Fron10der (changes requested):
just a one recommendation to fix

OleziO (approved):

PR #2202, opened 2023-04-28
Add solution and some new test cases

ruslanpashkov:


DidenkoDenis306 (approved):
🚀🚀🚀

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  if (typeof robot !== 'object' || robot === null || Array.isArray(robot)) {
+    throw new Error('Input must be an object');
+  }
```
naziamoff, line 12: what does array.isArray() will return if we pass plain object there?

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+
+  const inverted = {};
```
naziamoff, line 14: inverted what?

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+
+  for (const key in robot) {
+    if (inverted[robot[key]]) {
+      return null;
```
naziamoff, line 18: do you really need to return null here? I believe `break` or `continue` would work better here, depending on your need

naziamoff (changes requested):
not bad, but see some suggestions

src/inverseRobot.js
ruslanpashkov: @naziamoff [If any of the object values are repeated, return 'null'.](https://github.com/mate-academy/js_inverse_robot/blob/master/readme.md#:~:text=If%20any%20of%20the%20object%20values%20are%20repeated%2C%20return%20%27null%27.)

src/inverseRobot.js
naziamoff: Accordingly to the [thread](https://mateacademy.slack.com/archives/C050SHYFFR9/p1683025123681399) - I agree with your approach, this now makes sense to me 👍

naziamoff (approved):
yup, I shouldn't have rejected this one. Great job 👍

PR #2204, opened 2023-05-01
Solution 1

denys2:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const inverted = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (inverted[value] !== undefined) {
+      // if a value is repeated, return null
```
witflash, line 16: There is no need to leave such comments.

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+      return null;
+    }
+    inverted[value] = key;
```
witflash, line 19: It's better to do it vice versa:
```
if (inverted[value] === undefined) {
  inverted[value] = key;
}
```
So you can avoid the additional return statement.

witflash (approved):
Well done!
Pay attention to my comments. It may improve your future work.

PR #2205, opened 2023-05-01
add task solution

maksym-nemera:


DidenkoDenis306 (approved):
Good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const value in robot) {
+    const keys = robot[value];
```
naziamoff, line 13: This confused me a little. This is a single value, but from its name it seems it's in plural.
```suggestion
    const key = robot[value];
```

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+    if (Object.keys(newRobot).includes(keys)) {
+      return null;
+    }
```
naziamoff, line 17: not bad, but same as in my comment above:
```suggestion
    if (Object.keys(newRobot).includes(key)) {
      return null;
    }
```

naziamoff (approved):
Looks good 👍

PR #2206, opened 2023-05-01
inverseRobot solution

DidenkoDenis306:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotKeys = Object.keys(robot).reverse();
+  const robotValues = Object.values(robot).reverse();
+  const invRobot = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
```
33Orange, line 14: Instead of using separate arrays for keys and values, use a 'for...in' loop to iterate through the properties of the 'robot' object.

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotKeys = Object.keys(robot).reverse();
+  const robotValues = Object.values(robot).reverse();
```
33Orange, line 11: You don't need to use reverse. Both times you flip arrays for no reason

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const invRobot = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
+    if (invRobot.hasOwnProperty(robotValues[i])) {
+      return null;
+    }
+
+    invRobot[robotValues[i]] = robotKeys[i];
```
33Orange, line 19: You can take the value inside the iteration and not create 2 extra arrays.

33Orange (changes requested):
Use `for .. in` loop to iterate through the props of the "robot" object instead of using 2 separated arrays of keys and values.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invRobot = {};
```
Anna-Tkachenko, line 10: Don't make the variable name shortened. Let's use full variable names

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2210, opened 2023-05-03
Solution

Oleksii-Hrytsenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
```
VeronikaSynieokaia, line 9: Consider adding a more descriptive parameter name, such as 'inputRobot', to make the code more readable.

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
-  // write code here
+  const result = {};
```
VeronikaSynieokaia, line 10: Consider renaming the 'result' variable to 'invertedRobot' for better readability and understanding of the code.

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  for (const robotKey in robot) {
```
VeronikaSynieokaia, line 12: It is recommended to use 'let' instead of 'const' for the 'robotKey' variable since it can change in each iteration of the loop.

VeronikaSynieokaia (changes requested):
some fixes are required

src/inverseRobot.js
Oleksii-Hrytsenko: The message "'robotKey' is never reassigned. Use 'const' instead. " is shown in VS Code.
After starting the "npm test" command in terminal the "let" is changed to "const" automatically.

src/inverseRobot.js
Oleksii-Hrytsenko: fixed

src/inverseRobot.js
Oleksii-Hrytsenko: fixed

Ivan-Shvorak (approved):
Everything is okay

PR #2214, opened 2023-05-04
Inverse Robot

levchenko-dmytro:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const mirror = {};
+
+  for (const part in robot) {
+    if (!mirror[robot[part]]) {
+      mirror[robot[part]] = part;
+    } else {
```
vlad-demchuk, line 15:
```suggestion
    if (mirror[robot[part]]) {
      return null
    } 
    
     mirror[robot[part]] = part;
```
to avoid `else`

vlad-demchuk: also use `hasOwnProperty` method

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #2218, opened 2023-05-06
create

LeoSpartaDeV:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const resObj = {};
+  const keyForObj = Object.values(robot);
+  const valueForObj = Object.keys(robot);
```
SanyaBratashchuk, line 12: Simplify the solution. Object.values and Object.keys is redundant here

SanyaBratashchuk (changes requested):

OleziO (approved):

PR #2219, opened 2023-05-06
make inverse

pos1bl:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]] !== undefined) {
```
Volodymyr-Melnyk-Fron10der, line 13: Consider using 'Object.prototype.hasOwnProperty.call()' to check for properties directly on the object, rather than checking for undefined values.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+      return null;
+    }
+
+    newRobot[robot[key]] = key;
```
Volodymyr-Melnyk-Fron10der, line 17: Use object destructuring to improve readability when assigning values from 'robot'.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
```
Volodymyr-Melnyk-Fron10der, line 12: optional | consider using Object.entries()

Volodymyr-Melnyk-Fron10der (changes requested):

vlad-demchuk (approved):

PR #2220, opened 2023-05-06
Develop

GeorgeSavostikov:


src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  let newRobot = {};
```
TarasHoliuk, line 10:
```suggestion
  const newRobot = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+  const allRobotValues = Object.values(robot);
+
+  for (let i = 0; i < allRobotValues.length; i++) {
```
TarasHoliuk, line 14: your solution is too complicated (nested loops are not the best practice in case you can get rid of them)

the last loop (on the 24-th row) is good enough, just add a check there: if  `newRobot[robot[key]]` is defined return `null` immediately

TarasHoliuk (changes requested):

OleziO (approved):

PR #2223, opened 2023-05-07
Solution

bodyabandura:


src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
+    if (result.hasOwnProperty(robot[key])) {
```
sergii-nosachenko, line 30: Replace 'result.hasOwnProperty()' with 'Object.prototype.hasOwnProperty.call()' to prevent potential issues with overridden 'hasOwnProperty'

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  for (const key in robot) {
```
sergii-nosachenko, line 29: It's better to use the Object.keys() or Object.values() methods to iterate over 'robot' keys (and values) to avoid potential issues with the prototype chain

src/inverseRobot.js
```diff
@@ -24,7 +24,17 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
```
sergii-nosachenko, line 27: Use the descriptive name, f.e. `inversedRobot`

sergii-nosachenko (changes requested):

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
...
+  const inversedRobot = {};
+
+  const keys = Object.keys(robot);
+
+  for (const key of keys) {
+    if (Object.prototype.hasOwnProperty.call(inversedRobot, robot[key])) {
+      return null;
+    } else {
+      inversedRobot[robot[key]] = key;
+    }
```
sergii-nosachenko, line 36:
```suggestion
    }
    
    inversedRobot[robot[key]] = key;
```

If you use the `return` keyword within an `if` statement, there is no need to include the `else` branch since it will never be reached.

src/inverseRobot.js
```diff
@@ -24,7 +24,19 @@
  * @return {object}
  */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  const keys = Object.keys(robot);
+
+  for (const key of keys) {
```
sergii-nosachenko, line 31:
```suggestion
  for (const [key, value] of Object.entries(robot)) {
```
By utilizing the `Object.entries()` method, you can conveniently access both the key and value simultaneously.

sergii-nosachenko (approved):
I have left some comments on how you can improve the solution.

PR #2229, opened 2023-05-08
Develop

okosohor:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
```
Anna-Tkachenko, line 10: Let's create more descriptive variable name, for now it's not clear what should be stored in this variable.

Anna-Tkachenko (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
```
Anna-Tkachenko, line 10:
```suggestion
  const reversedRobot = {};
```

Anna-Tkachenko (approved):
⭐️

PR #2230, opened 2023-05-08
add task solution

AlexKherson:


amashutikov (approved):
Good job, i've made it in a bit different way. In Checklist №3 said that we should pick better one name for result value.

silchyn (changes requested):
Pay attention to the comment above

src/inverseRobot.js
```diff
@@ -8,6 +8,22 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+  const valueList = [];
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (valueList.includes(value)) {
```
silchyn, line 17: where do we already have something like `valueList`? there is faster solution

silchyn (changes requested):

silchyn (approved):

PR #2235, opened 2023-05-08
solution

Yuliya1907:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invented = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (invented[value] !== undefined) {
+      return null;
+    }
```
IrynaBalandiukh, line 15: Please, check point 2 in the checklist

IrynaBalandiukh (changes requested):
Good job, but let`s fix the code style

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invented = {};
```
Anna-Tkachenko, line 10: It's not clear what should be stored in this variable. Let's think more about meaningful and clear variable name ⏰

Anna-Tkachenko (changes requested):

33Orange (commented):
Well done!

33Orange (approved):

PR #2238, opened 2023-05-09
Solution

BogdanParubchenk01:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
...
+  const result = {};
+  const check = Object.values(robot);
+
+  for (let i = 0; i < check.length; i++) {
+    if (check.indexOf(check[i]) !== i) {
+      return null;
+    }
+  }
+
+  for (const key in robot) {
```
anstsot, line 19: you can simplify the code and use one `for` loop. just add the check for second loop and return `null` if needed

anstsot (changes requested):
It’s almost done! Let’s improve a little

PR #2243, opened 2023-05-09
solution

thienlaoo:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
+
+    if (newRobot[newKey]) {
```
annholovko, line 15: [CODE STYLE] - use correct check for property presence in object. Some properties can be present, but still contain falsy value.

GOOD EXAMPLE:

if (key in robot) {
EVEN BETTER EXAMPLE:

BAD EXAMPLE:

if (robot[key] === undefined) {

if (robot[key]) {

annholovko (changes requested):
you need to write the correct check for property in condition

Anna-Tkachenko (approved):
Good job! ⭐️

PR #2244, opened 2023-05-09
Solution

kirilshorin:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
```
maksym-bobryk, line 10: it would be better to rename variable for a better intuitive understanding

maksym-bobryk (approved):
Well done🚀

src/inverseRobot.js
33Orange, line 10: Consider using a more descriptive name for the new object, like 'invertedRobot' instead of 'obj'.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+
+  for (const item in robot) {
+    if (obj[robot[item]] !== undefined) {
```
33Orange, line 13: Would be better to use obj.hasOwnProperty here

33Orange (changes requested):
Good job.

Some comments:
- Use proper names for variables.
- Use object.hasOwnProperty to check if property exist.

silchyn (approved):

PR #2245, opened 2023-05-09
Add task solution

IrynaBalandiukh:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    const robotKey = robot[key];
```
EugeneSherepa, line 13: Seems like it doesnt have any scene to use this variable, it would only increase the time of your page loading☕

EugeneSherepa (commented):
Everything but this variable is nice, change this for perfection💚

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2246, opened 2023-05-09
Solution

julshvets:


MaksimKad (approved):
GoodJ!

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const valuesOfRobotKeys = Object.keys(robot);
+  const robotKeys = Object.values(robot);
+  const reparedRobot = {};
```
33Orange, line 12: Consider renaming 'reparedRobot' to 'invertedRobot' for better clarity.

33Orange (changes requested):
Good job!

A few moments:
- Try to use more descriptive variable names
- Let's try solve this task without `Object.values` and `Object.keys`. You can iterate over an object with the `for in` operator. It would be better than create 2 arrays of keys and values.

silchyn (approved):

PR #2249, opened 2023-05-09
inverseRobot solution

panch1811:


ivankhrsh (approved):

src/inverseRobot.js
```diff
@@ -8,6 +8,16 @@
 
 function inverseRobot(robot) {
   // write code here
-}
+  const inverse = {};
+
+  for (const key in robot) {
+    if (Object.prototype.hasOwnProperty.call(inverse, robot[key])) {
```
33Orange, line 14: Consider using 'hasOwnProperty' on the 'inverse' object directly instead of using 'Object.prototype.hasOwnProperty.call'.

33Orange (changes requested):
Good job!

A few moments: 
- Use names that better describe the variable
- Consider using 'hasOwnProperty' on the 'inverse' object directly instead of using 'Object.prototype.hasOwnProperty.call'.

src/inverseRobot.js
```diff
@@ -8,6 +8,16 @@
...
   // write code here
-}
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (invertedRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
 
+    invertedRobot[robot[key]] = key;
```
silchyn, line 18:
```suggestion
    const value = robot[key];
    
    if (invertedRobot.hasOwnProperty(value)) {
      return null;
    }

    invertedRobot[value] = key;
```

silchyn (approved):

PR #2250, opened 2023-05-09
Solution

olehmahera:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
 function inverseRobot(robot) {
-  // write code here
-}
+  const copy = {};
+  const values = Object.values(robot);
+
+  for (const key in robot) {
+    const val = robot[key];
 
+    if (values.indexOf(val) !== values.lastIndexOf(val)) {
```
ulositska, line 16: Not the most obvious, but very interesting way to check for duplicates! 🎆

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const copy = {};
```
ulositska, line 10: It is technically a copy, but I'd suggest using a more descriptive name for a variable 
(Checklist 3)

ulositska (approved):
Especially appreciated how you've used a variable inside the 'for loop':fire::fire::fire:

src/inverseRobot.js
silchyn, line 16: try to find a more straightforward and faster solution

silchyn (changes requested):

33Orange (approved):
Well done!

silchyn (approved):

PR #2251, opened 2023-05-09
Solution

MaksimKad:


vchervinko (approved):
Nice)

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  let reversedObject = {};
```
Anna-Tkachenko, line 10: Don't use data type names in variable names

Anna-Tkachenko (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversed = {};
```
Anna-Tkachenko, line 10: It's not clear what should be stored in this variable
`reversedRobot` will be better variable name

Anna-Tkachenko (approved):
⭐️

PR #2253, opened 2023-05-09
Solution

mkorulczyk:


src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+  for (const key in robot) {
+    const value = robot[key];
+    if (inverse[value] !== undefined) {
```
DorotaLeniecDev, line 13: This would be more suitable here
```suggestion
    if (inverse.hasOwnProperty(value)) {
```

DorotaLeniecDev (changes requested):
Good job, one small change needed

PR #2256, opened 2023-05-09
SolutionInverseRobots

olena-bielozorova:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const expectRobot = {};
```
Ivan-Shvorak, line 10: it would be better to name it as expectedRobot

Ivan-Shvorak (approved):
Everything is okay but please check the comment

PR #2257, opened 2023-05-09
Solution

bohdan-mykhailenko:


SofiaLiubchenko (approved):
good job

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (newRobot[value]) {
```
33Orange, line 13: Change the condition to check if the property exists in the object, use 'hasOwnProperty' method.

33Orange (changes requested):
Good job!

Just one moment. Let's change the condition in the `if` statement using the object's hasOwnProperty method

silchyn (approved):

PR #2264, opened 2023-05-09
Solution

dimoshen:


growingananas (approved):
Great solution, though naming is a bit confusing)

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const normalRobot = {};
```
silchyn, line 10: what makes it normal?

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  for (const fakeKey in robot) {
```
silchyn, line 12: what makes it fake?

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+    const trueKey = robot[fakeKey];
```
silchyn, line 13: what makes it true?

silchyn (changes requested):

src/inverseRobot.js
dimoshen: names changed

src/inverseRobot.js
dimoshen: names changed

src/inverseRobot.js
dimoshen: names changed

dinaminahi (approved):
👍

silchyn (approved):

PR #2265, opened 2023-05-09
made js inverse robot

vladyslav-lapshyn:


serentinos (approved):
Good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inverseObject = {};
+
+  for (const key in robot) {
+    if (inverseObject.hasOwnProperty(robot[key])) {
+      return null;
+    }
+    inverseObject[robot[key]] = key;
+  }
```
33Orange, line 17: Value `robot[key]` was used 2 times.
Maybe it's better to put it in a variable?

33Orange (changes requested):
Well done ! 🔥 

Need to fix just one moment with readability.

src/inverseRobot.js
vladyslav-lapshyn: @33Orange thanks, it's good advice

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverseObject = {};
```
Anna-Tkachenko, line 10: Don't use data type name for variable names

Anna-Tkachenko (approved):

PR #2267, opened 2023-05-09
Solution

SofiaLiubchenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const noTranspositionRobot = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
Anna-Tkachenko, line 13: It's not clear for what purpose you've added this check
In `for` loop we have only existing robot keys, so this check will always return `true`

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #2269, opened 2023-05-09
Solution

maryanmats:


src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inverse = {};
+
+  for (const key in robot) {
```
33Orange, line 13: Consider using 'let' instead of 'const' for variable 'key' since the value may change in each iteration.

src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
...
+    const value = robot[key];
```
33Orange, line 14: Consider using 'let' instead of 'const' for variable 'value' since the value may change in each iteration.

33Orange (changes requested):
Well done ! 🔥 

Just one moment with `inverse` variable name. 
Let's change it to something more descriptive. (for example: `invertedRobot`)

DidenkoDenis306 (approved):

naziamoff (approved):
👍

PR #2271, opened 2023-05-10
Solution

Yuliia-kucheriavenko:


src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const rightRobot = {};
+
+  for (const key in robot) {
+    if (rightRobot.hasOwnProperty(robot[key])) {
```
33Orange, line 14: Use 'Object.prototype.hasOwnProperty.call' instead of calling 'hasOwnProperty' directly on the object to avoid potential issues with overwritten properties.

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
```
33Orange, line 9: Consider using arrow function syntax for 'inverseRobot' function.

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
...
   // write code here
+  const rightRobot = {};
```
33Orange, line 11: Consider using a more descriptive name for 'rightRobot', such as 'invertedRobot'.

33Orange (changes requested):
Good job ! 🔥 

Just a few moments:.
- Let's change variable name to something more descriptive. (for example: invertedRobot)
- You use value robot[key] 2 times. Maybe would be better save it to variable?

Anna-Tkachenko (approved):
⭐️

PR #2273, opened 2023-05-10
add solution

growingananas:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
33Orange, line 10: Let's change variable name to something more descriptive. (for example: `invertedRobot`)

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (value in newRobot) {
```
33Orange, line 15: Use method `.hasOwnProperty` to check if such a property is in an object

33Orange (changes requested):
Good job ! 🔥

Just a few moments:.
- Let's change variable name to something more descriptive. (for example: `invertedRobot`)
- Use method `.hasOwnProperty` to check if such a property is in an object

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (invertedRobot.hasOwnProperty([value])) {
```
silchyn, line 15:
```suggestion
    if (invertedRobot.hasOwnProperty(value)) {
```

silchyn (approved):

PR #2285, opened 2023-05-14
Solution

Lidia-Pintos-Medina:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (newRobot[value]) {
```
Anastasiia-Svintsova, line 15: Change the condition to check if the property exists in newRobot. Use `newRobot.hasOwnProperty(value)` instead of 'if (newRobot[value])'.

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):
👍

PR #2288, opened 2023-05-16
add solution

dimarogkov:


src/inverseRobot.js
```diff
@@ -1,13 +1,29 @@
...
+    (item, index) => valuesArr.indexOf(item) !== index
+  );
+
+  if (isValuesRepeated) {
+    return null;
+  }
+
+  for (const key in robot) {
+    res[robot[key]] = key;
+  }
```
IvanFesenko, line 24: No need to do extra loops iterations. You can check it in your for loop
```suggestion
  for (const key in robot) {
    const newKey = robot[key];
    
    if (res.hasOwnProperty(newKey )) {
      return null;
    }
     
    res[newKey] = key;
  }
```

IvanFesenko (changes requested):
Good job, lets improve your code

IvanFesenko (approved):

PR #2293, opened 2023-05-17
Solution

a-kulaev:


src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+  const robotValues = Object.values(robot);
+  const newObject = {};
+
+  for (let i = 0; i < robotValues.length; i++) {
+    for (let j = i + 1; j < robotValues.length; j++) {
+      if (robotValues[i] === robotValues[j]) {
+        return null;
+      }
+    }
+  }
```
Anastasiia-Svintsova, line 20: The loop is redundant actually. Just use .hasOwnProperty() method inside the for ... in loop

src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+
+  for (const i in robotKeys) {
```
Anastasiia-Svintsova, line 22:
```suggestion
  for (const key in robot) {
```

Anastasiia-Svintsova (changes requested):
Hey! Simplify the solution, please. Use my comments as a hint

src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+  const newObject = {};
+
+  const uniqueValues = {};
+
+  for (let i = 0; i < robotValues.length; i++) {
+    if (uniqueValues.hasOwnProperty(robotValues[i])) {
+      return null;
+    }
+
+    uniqueValues[robotValues[i]] = true;
```
polosanya, line 20: This logic is actually redundant

src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
...
+  }
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    newObject[value] = key;
```
polosanya, line 26: Just check here if value is not present in new robot

polosanya (changes requested):
Almost done! Consider using only one loop to complete this task

src/inverseRobot.js
a-kulaev: Thank You. I think I got it right. Check please

tetiana-ivanchenko (approved):

PR #2294, opened 2023-05-18
Solution

IraPolchanin:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const prop in robot) {
```
volodymyr-soltys97, line 12: Consider using more meaningful variable name instead of 'prop', such as 'key' or 'property'.

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
volodymyr-soltys97, line 10: Consider using a more descriptive name for the 'newRobot' variable, such as 'invertedRobot'.

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+
+  for (const prop in robot) {
+    if (newRobot.hasOwnProperty(robot[prop])) {
+      return null;
+    }
+    newRobot[robot[prop]] = prop;
```
volodymyr-soltys97, line 16: Also you can create a variable for `robot[prop]` and use this variable

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more sophisticated

polosanya (approved):
Nice!

PR #2300, opened 2023-05-20
Solution

hulchak:


src/inverseRobot.js
```diff
@@ -4,10 +4,23 @@
...
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const result = {};
+
+  const keysRobot = Object.values(robot);
```
TarasHoliuk, line 12: The name 'keysRobot' is misleading, it should be 'valuesRobot' as it stores object values.

TarasHoliuk: I understand why you named it `keys` - it will be the keys in the future - but you might show it in the name as well. Anyway not critical here - approved

TarasHoliuk (approved):

PR #2303, opened 2023-05-22
Solution

superzuc:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
-  // write code here
+  const invertedRobot = {};
+  const values = Object.values(robot);
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (values.indexOf(value) !== values.lastIndexOf(value)) {
+      return null;
+    }
```
anstsot, line 18: better to use `hasOwnProperty`

anstsot (changes requested):
Good job, but there is one fix to improve your code.

PR #2309, opened 2023-05-26
inverseRobot was created

EkaterinaBessonnaya:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const rightRobot = {};
```
SanyaBratashchuk, line 10:
```suggestion
  const inversedRobot = {};
```

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+      return null;
+    }
+  }
+
+  return rightRobot;
 }
 
 module.exports = inverseRobot;
+
+// console.log(inverseRobot(rightRobot));
```
SanyaBratashchuk, line 25: do not leave comments

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const rightRobot = {};
+
+  for (const key in robot) {
+    if (!rightRobot[robot[key]]) {
+      rightRobot[robot[key]] = key;
+    } else if (rightRobot) {
+      return null;
+    }
```
SanyaBratashchuk, line 17:
```suggestion
    } else {
      return null;
    }
```

SanyaBratashchuk (changes requested):

src/inverseRobot.js
EkaterinaBessonnaya: I can't change variable on this name, cause fall linter in VSC.
 I will leave the old one.

SanyaBratashchuk (approved):

DarkMistyRoom (approved):

polosanya (approved):

PR #2311, opened 2023-05-29
Solution

s3rg4sh:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+  const usedKeywords = [];
+
+  for (const key in robot) {
+    if (!usedKeywords.includes(key) && !usedKeywords.includes(robot[key])) {
```
etojeDenys, line 14: we can only check if `inverseRobot.hasOwnProperty(robot[key])`

src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+  const usedKeywords = [];
```
etojeDenys, line 11: redundant array

etojeDenys (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+  // const usedKeywords = [];
```
mykhalenych, line 11: delete all comments before commit

mykhalenych (approved):

PR #2312, opened 2023-05-29
Solution

Ilona1697:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const exchangedKeyValueObj = {};
+
+  for (const key in robot) {
```
mgrinko, line 12: you could use `Object.entries` to iterate only the own properties
```suggestion
  for (const [key, value] of Object.entries(robot)) {
```

mgrinko (approved):

PR #2319, opened 2023-05-31
add inverse

Misha408:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+  const keys = Object.keys(robot);
```
vlad-demchuk, line 11: you can reach the same using one for...in loop

vlad-demchuk: Object.values and Object.keys are the same that using for...in twice

vlad-demchuk (changes requested):

nazarmatsevych (approved):

nazarmatsevych (approved):

PR #2320, opened 2023-06-01
solution

NataliMax:


src/inverseRobot.js
```diff
@@ -6,8 +6,25 @@
  * @return {object}
 */
 
-function inverseRobot(robot) {
-  // write code here
+function inverseRobot(oldRobot) {
+  const newRobot = {};
+  let newRobotValue = '';
+  let newRobotKey = '';
```
etojeDenys, line 12: The variable 'newRobotKey' is not necessary, you can directly use 'oldRobot[oldKey]' in the loop.

src/inverseRobot.js
```diff
@@ -6,8 +6,25 @@
...
+  const oldRobotValues = [];
+
+  for (const oldKey in oldRobot) {
+    for (const oldValue of oldRobotValues) {
+      if (oldValue === oldRobot[oldKey]) {
+        return null;
+      }
```
etojeDenys, line 19: we can simplify logic
```
if(newRobot[oldRobot[oldKey]]) {
  return null;
}
```

src/inverseRobot.js
```diff
@@ -6,8 +6,25 @@
...
+    }
+    newRobotValue = oldKey;
+    newRobotKey = oldRobot[oldKey];
+    newRobot[newRobotKey] = newRobotValue;
+    oldRobotValues.push(oldRobot[oldKey]);
```
etojeDenys, line 24: same here
```
newRobot[oldRobot[oldKey]] = oldKey
```

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #2327, opened 2023-06-04
solution

hrynkevych:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const ROBOT_VALUES = Object.values(robot);
+
+  function findDuplicates(arry) {
+    return arry.filter((item, index) => arry.indexOf(item) !== index);
+  };
+
+  const DUP_VALUES = findDuplicates(ROBOT_VALUES);
```
mykhalenych, line 16: Use 'const duplicateValues' instead of 'const DUP_VALUES' to follow camelCase naming conventions.

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const ROBOT_VALUES = Object.values(robot);
+
+  function findDuplicates(arry) {
```
mykhalenych, line 12: The 'findDuplicates' function should be defined outside of the 'inverseRobot' function to avoid redefining it every time the 'inverseRobot' function is called, also change arry to array

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+    return arry.filter((item, index) => arry.indexOf(item) !== index);
+  };
+
+  const DUP_VALUES = findDuplicates(ROBOT_VALUES);
+
+  if (DUP_VALUES.length >= 1) {
```
mykhalenych, line 18: Instead of checking if 'DUP_VALUES.length >= 1', you can use 'if (DUP_VALUES.length)' which will evaluate to 'true' if there are any duplicates.

mykhalenych (approved):

PR #2329, opened 2023-06-05
Solution

IrynaMP:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotCorrectedObject = {};
+  const swappedValues = {};
```
mykhalenych, line 11: Rename 'swappedValues' to 'invertedRobot' to better represent its purpose.

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotCorrectedObject = {};
```
mykhalenych, line 10: The variable 'robotCorrectedObject' is not necessary. You can use 'swappedValues' to store the final result.

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
+      const value = robot[key];
+
+      if (swappedValues.hasOwnProperty(value)) {
+        return null;
+      }
+
+      robotCorrectedObject[value] = key;
```
mykhalenych, line 21: Instead of using 'swappedValues[value] = true;', update the 'invertedRobot' directly, i.e., 'invertedRobot[value] = key;'.

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotCorrectedObject = {};
+  const swappedValues = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
mykhalenych, line 14: The 'hasOwnProperty' check is not necessary, as using 'for...in' already checks for properties in the object.

mykhalenych (changes requested):

OleziO (approved):

PR #2341, opened 2023-06-08
Solution

yehor-kor:


src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+  for (const key in robot) {
+    keys.push(robot[key]);
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    if (keys.includes(keys[i], i + 1)) {
+      return null;
+    }
+  }
```
agniya-i, line 20: Instead of adding extra logic to find repeated values, you can do this right in the loop where you inverse, just add a check if `inverse` object already has such property

agniya-i (changes requested):

src/inverseRobot.js
yehor-kor: Okay, I know how to fix that

polosanya (approved):
Nice!

PR #2343, opened 2023-06-09
Solution

MarianVolytskyi:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
```
tetiana-ivanchenko, line 13: better to move robot[key] in the variable, as you use it twice

tetiana-ivanchenko (approved):

PR #2344, opened 2023-06-09
Solution

IraShtynda:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
```
irinabolbat, line 12: You can use `Object.entries` to find `key` and `value`

irinabolbat (changes requested):
Good job 👏 
Please update your code

irinabolbat (approved):
👍

PR #2346, opened 2023-06-09
Solution

dianadiv:


ollavka (approved):
Excellent)

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
```
tetiana-ivanchenko, line 12: actually, you can use 
` for (const key in robot) {`

tetiana-ivanchenko (approved):

PR #2349, opened 2023-06-09
Finished inverseRobot task

IgorOksentyuk:


src/inverseRobot.js
```diff
@@ -7,7 +7,15 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const entries = Object.entries(robot).map(([key, value]) => [value, key]);
+  const allElements = entries.flatMap(arr => [...arr]);
+  const filteredArray = [...new Set(allElements)];
```
tetiana-ivanchenko, line 12:
```suggestion
  const values = Object.values(robot);
  const filteredArray = [...new Set(values)];
```

you can simplify your solution

tetiana-ivanchenko (changes requested):

anstsot (approved):

PR #2357, opened 2023-06-10
Solution

anya-shults:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+  let robotRealKey;
+
+  for (const key in robot) {
+    robotRealKey = robot[key];
```
tetiana-ivanchenko, line 14: if you use variable only inside loop, you should define it here

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2358, opened 2023-06-10
add task solution

kbekher:


src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newRobot = { };
+  const keys = Object.keys(robot);
+
+  for (let i = 0; i < keys.length; i++) {
```
etojeDenys, line 14: Using two nested loops for checking if any values are repeated is not efficient. Instead, you can use a single loop and check if the value already exists in the newRobot object before adding it.

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #2359, opened 2023-06-10
added task solution

tetianakostiuk1:


fawnique (approved):
Great job! Well done :)

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversed = {};
```
eleonora-demikh, line 10: You could also not create a new object, but change the old one, because the task asks to "repair" it, not create another

eleonora-demikh (approved):
Well done!

src/inverseRobot.js
tetiana-ivanchenko: > You could also not create a new object, but change the old one, because the task asks to "repair" it, not create another

image

tetiana-ivanchenko (approved):

PR #2360, opened 2023-06-10
task solved

LinaDo1609:


src/inverseRobot.js
```diff
@@ -4,10 +4,25 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.values(robot);
+  const values = Object.keys(robot);
+  let value = '';
+  let key = '';
+  const result = {};
+
+  for (let i = 0; i < keys.length; i++) {
+    if (keys.slice(i + 1).includes(keys[i])) {
```
OleziO, line 17: At the last iteration, i + 1 will be out of arr

OleziO (changes requested):

src/inverseRobot.js
```diff
@@ -4,10 +4,25 @@
...
+      return null;
+    }
+    key = keys[i];
+    value = values[i];
+    result[key] = value;
+  }
+
+  return result;
```
SanyaBratashchuk, line 25:
```suggestion
  const result = {};

  for (let key in robot) {
    if (robot[key] in result) {
      return null;
    }
      result[robot[key]] = key
  }

  return result;
```

SanyaBratashchuk: Simplify solution

SanyaBratashchuk (changes requested):

src/inverseRobot.js
```diff
@@ -1,13 +1,20 @@
-'use strict';
+"use strict";
 
 /*
```
OleziO, line 3: Remove all comments from the code.

OleziO (changes requested):

VitaliyBodnar (approved):
Well done

PR #2362, opened 2023-06-10
Solution

fawnique:


outofs (approved):
Nice solution! Aprove! I had similar way to resolve this.

tetiana-ivanchenko (approved):

PR #2363, opened 2023-06-11
implement inverse Robot

kochetov-yevhen-may29:


anstsot (changes requested):
Good start, but there are few improvements:
1. you can use `for ... in` instead of raw for loop
2. there is no need to use 2 loops, rewrite to use only one
3. you can check is the inversed robot has the same value directly in `for ... in` loop and return `null` right there. no need to do job twice

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const INVERSE_KEYS_AND_VALUES = {};
```
And678, line 10: CAPITAL_SNAKE_CASE should be used only for global constants that will never mutate. (e.g. `const PI = 3.14151...`).

Here you should regularPascalCase. Check out [mates naming convention](https://github.com/mate-academy/style-guides/blob/master/javascript.md#23-naming-conventions).

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+
+  for (const key in robot) {
+    if (INVERSE_KEYS_AND_VALUES[robot[key]] !== undefined) {
```
And678, line 13: How about this object, `const robot = {"a": 123, "b": undefined};`, it will fail your function. Read the second point of [checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md)

And678 (changes requested):
Good work, but some things can be improved

And678 (approved):
:+1: :+1:

PR #2364, opened 2023-06-11
Task done

AntonBaranskyi:


src/inverseRobot.js
```diff
@@ -4,10 +4,34 @@
  * @param {object} robot
  *
  * @return {object}
-*/
-
+ */
 function inverseRobot(robot) {
-  // write code here
+  const keys = Object.values(robot);
+  const obj = {};
```
zoriana-melnychuk, line 10: Consider using more descriptive names for the variables(for example valueCounts instead of obj). Same for the rows below

zoriana-melnychuk (changes requested):

src/inverseRobot.js
```diff
@@ -4,10 +4,34 @@
...
+    }
+  }
+
+  const robotValues = Object.values(countObj).some((num) => num > 1);
+
+  if (robotValues) {
+    return null;
+  }
+
+  const invertObj = Object.fromEntries(
```
juljaVysotska, line 28: The current implementation only reverses the object entries and maps them, which does not guarantee the correct order of keys and values. Use a simpler approach with a new object and a 'for ... in' loop to swap keys and values.

src/inverseRobot.js
```diff
@@ -4,10 +4,34 @@
...
  * @return {object}
-*/
-
+ */
 function inverseRobot(robot) {
-  // write code here
+  const robotKeys = Object.values(robot);
+  const countObj = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
```
juljaVysotska, line 12: Use 'for ... of' loop instead of a standard 'for' loop for better readability.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them!

anstsot (changes requested):
Almost done but your approach is overcomplicated. you can just use `for ... of` loop and check right there if the values already exist in your reversed robot
let's simplify the solution

mgrinko (approved):

PR #2367, opened 2023-06-12
inverseRobot solution

ivanbilotserkivskiy:


ArtemAndrushchenko (approved):
Good job

anstsot (changes requested):
Good start, but there are few improvements:
1. you can use `for ... in` instead of raw for loop
2. there is no need to use 2 loops, rewrite to use only one
3. you can check is the inversed robot has the same value directly in `for ... in` loop and return `null` right there. no need to do the job twice

And678 (approved):

PR #2368, opened 2023-06-12
Solution

ShamievDavid:
I lessons mentor said to use new 'Object.hasOwn' instead of 'obj.hasOwnProperty' but it is only available since node.js v16.9.0.

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const swappedObj = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
+      if (swappedObj.hasOwnProperty(robot[key])) {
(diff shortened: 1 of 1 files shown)


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const swappedObj = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
VitaliyBondarenko1982, line 13:
```suggestion
```
looks like redundant this if

VitaliyBondarenko1982 (changes requested):

src/inverseRobot.js
ShamievDavid: you are right.

VitaliyBondarenko1982 (approved):

PR #2371, opened 2023-06-12
add task solution

ArtemAndrushchenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (invertedRobot[value] !== undefined) {
+      return null;
+    }
```
OlhaSheliakina, line 17: Nice work!
Just one little thing to change
Please see checklist item No.2. It would be better to check for duplicates using hasOwnProperty() method
js-checklist-inverseRobot

OlhaSheliakina (changes requested):

src/inverseRobot.js
ArtemAndrushchenko: ty for your help !

PR #2373, opened 2023-06-12
Add task solution

vladmakohon:


AlyonaSarapina (approved):
good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (Object.hasOwnProperty.call(fixedRobot, robot[key])) {
```
juljaVysotska, line 13: Use 'fixedRobot.hasOwnProperty' instead of 'Object.hasOwnProperty.call' for better readability

juljaVysotska (changes requested):
Good job! I found some isues, please, check and fix it.

And678 (changes requested):
Ignore this, table sync

src/inverseRobot.js
vladmakohon: Sure. However, isn't it better to prevent a scenario when a built-in method is overridden?

And678 (approved):

Considering your question, nice thinking :brain: , indeed it can be overwritten. E.g. in case when we have something like this: 
```js
const a = JSON.parse('{"hasOwnProperty": 123}');
```

In the current scope its ok to use the method directly as this is an exercise.

PR #2377, opened 2023-06-12
add task solution

AlyonaSarapina:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const keyToInverse in robot) {
+    if (inversedRobot.hasOwnProperty(robot[keyToInverse])) {
```
FlameInBrains, line 13: It is better to create a variable for robot[keyToInverse] since it is used more than one time.

FlameInBrains (changes requested):

src/inverseRobot.js
AlyonaSarapina: Good idea. I've implemented it.

FlameInBrains (approved):

And678 (approved):

PR #2378, opened 2023-06-12
Solution

NadiaKovalchuk:


And678 (approved):
Nice! Original solution, cool idea to use a Set in order to check for duplicate keys. Approve :heavy_check_mark:

PR #2382, opened 2023-06-12
Solution

kylyvnyk:


dipinchuk (approved):
Good job!
Small improvement: I think it is better to check for the same values ​​at the beginning, so as not to enter a loop if values are repeated
```
  const uniqueValues = [...new Set(values)];

  if (uniqueValues.length !== values.length) {
    return null;
  }
  ```

anstsot (approved):

PR #2384, opened 2023-06-12
Solution

olegpaliy:


taras-bilas (approved):
Good job!
It would be good for the future to name variables more semantically

src/inverseRobot.js
```diff
@@ -4,10 +4,19 @@
...
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const obj = {};
+
+  for (const key in robot) {
+    if (obj.hasOwnProperty([robot[key]])) {
```
tetiana-ivanchenko, line 13:
```suggestion
    if (obj.hasOwnProperty(robot[key])) {
```

also move `robot[key]` in the variable

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2388, opened 2023-06-12
add solution

pavlokuraksin:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  const robotKeys = Object.values(robot);
+  const robotValues = Object.keys(robot);
+
+  for (let index = 0; index < robotKeys.length; index++) {
```
tetiana-ivanchenko, line 14: use `for in` loop to iterate robot object, in that case robotKeys and robotValues are redundant

tetiana-ivanchenko (changes requested):

anstsot (approved):

PR #2394, opened 2023-06-13
Solution

Sirius9312:
Siltus inversus task

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -8,8 +8,8 @@
 
 function inverseRobot(robot) {
   // write code here
-  const ONE_VALUE_IN_OBJECT = 1;
-  let countsObj = {};
+  const INITIAL_COUNT = 1;
+  const countsObj = {};
 
   const resObj = {};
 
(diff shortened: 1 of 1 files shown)


src/inverseRobot.js
```diff
@@ -8,6 +8,36 @@
 
 function inverseRobot(robot) {
   // write code here
+  const ONE_VALUE_IN_OBJECT = 1;
```
volodymyr-soltys97, line 11: Consider using a more descriptive constant name such as 'INITIAL_COUNT' instead of 'ONE_VALUE_IN_OBJECT'

src/inverseRobot.js
```diff
@@ -8,6 +8,36 @@
...
+    if (countsObj[robot[key]] !== undefined) {
+      countsObj[robot[key]]++;
+    } else {
+      countsObj[robot[key]] = ONE_VALUE_IN_OBJECT;
+    }
+
+    countsObj = Object.assign(countsObj, countsObj[robot[key]]);
+  }
+
+  for (const key in countsObj) {
```
volodymyr-soltys97, line 30: The for loop at line 23 is unnecessary, you can already check for duplicate values and construct the result object in the first loop

src/inverseRobot.js
```diff
@@ -8,6 +8,36 @@
...
+  }
+
+  for (const key in robot) {
+    if (countsObj[robot[key]] !== undefined) {
+      countsObj[robot[key]]++;
+    } else {
+      countsObj[robot[key]] = ONE_VALUE_IN_OBJECT;
+    }
+
+    countsObj = Object.assign(countsObj, countsObj[robot[key]]);
```
volodymyr-soltys97, line 27: The usage of 'Object.assign' here is unnecessary and incorrect. It should be removed

src/inverseRobot.js
```diff
@@ -8,6 +8,36 @@
...
+  const ONE_VALUE_IN_OBJECT = 1;
+  let countsObj = {};
+
+  const resObj = {};
+
+  if (Object.keys(robot).length === 0) {
+    return {};
+  }
+
+  for (const key in robot) {
```
volodymyr-soltys97, line 20: It's better to use 'for (const [key, value] of Object.entries(robot))' for iterating through the object

src/inverseRobot.js
```diff
@@ -8,6 +8,36 @@
...
+    if (countsObj[robot[key]] !== undefined) {
```
volodymyr-soltys97, line 21: You can create a variable for `robot[key]`

volodymyr-soltys97 (changes requested):
Good job 👍 

Viktor-Kost (approved):

PR #2399, opened 2023-06-14
Solution

valet2102:


Vayts:
Цікаве рішення!

Vayts (approved):
good job!

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
 
 function inverseRobot(robot) {
   // write code here
+  const newObject = {};
+  const array = Object.entries(robot);
```
dffUqp, line 12: use more descriptive name.

src/inverseRobot.js
```diff
@@ -8,6 +8,17 @@
...
+
+  for (let i = 0; i < array.length; i++) {
```
dffUqp, line 14: let try to use for in loop to go through object

lear more about `for in`  - [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

src/inverseRobot.js
```diff
@@ -8,6 +8,41 @@
...
+  const INVERSE_ROBOT = {};
+
+  // #1 use for in
+
+  for (const key in robot) {
+    if (INVERSE_ROBOT[robot[key]]) {
+      return null;
+    }
+    INVERSE_ROBOT[robot[key]] = key;
+  }
```
BudnikOleksii, line 20: The naming convention for variables in javascript is camelCase. Only global constants can have names with current style.
Checklist: 
2. [CODE STYLE] - use correct check for property presence in object. Some properties can be present, but still contain falsy value.

BudnikOleksii (changes requested):
It works in this case, but a few things need to be fixed.
Also, remove comments.

dffUqp (approved):
GJ✔️

PR #2407, opened 2023-06-17
inverse_robot

NazarVasylyshyn:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const rezult = {};
```
volodymyr-soltys97, line 10: Consider using a more descriptive variable name instead of 'rezult'. For example, 'invertedRobot'.

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+
+  for (const key in robot) {
+    if (robot[key] in rezult) {
```
volodymyr-soltys97, line 13: Also you can create a variable for `robot[key]` and use it

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Awesome!

PR #2410, opened 2023-06-19
Solution

tarovyk:


src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
...
+  for (let i = 0; i < robotValues.length; i++) {
+    const curr = robotValues[i];
+    const remaining = robotValues.slice(i + 1);
+
+    if (remaining.includes(curr)) {
+      return null;
+    }
+  }
+
+  return Object.fromEntries(Object.entries(robot).map(a => a.reverse()));
```
loralevitska, line 22: Overall good work. Next time try to write more simpler code
Ex for this task:
```const result = {};

  for (const key in robot) {
    const newKey = robot[key];

    if (result.hasOwnProperty(newKey)) {
      return null;
    }

    result[newKey] = key;
  }

  return result;
  ```

loralevitska (approved):

PR #2415, opened 2023-06-20
Solution

petylep:


mykhalenych (changes requested):

etojeDenys (changes requested):
try to use `for in` as the mentor said it will simplify the logic

petylep:
> try to use `for in` as the mentor said it will simplify the logic

Use, but it really simplify the logic? I delete only one line...

demetrxx (approved):
Well done 👍 Go ahead!

PR #2418, opened 2023-06-21
add task solution

707Andrew:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const [key, value] of Object.entries(robot)) {
```
etojeDenys, line 12: Consider using 'for...in' loop for better readability

etojeDenys (approved):
lgtm

PR #2420, opened 2023-06-21
Solution

Constantin-Haranskiy:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const values = Object.values(robot);
+
+  if (values.length !== new Set(values).size) {
+    return null;
+  }
+
+  const resultObj = {};
+  const params = Object.entries(robot);
```
SanyaBratashchuk, line 17: Try to simplify solution Object.values, Set, entries are redundant here. All what you need is just a  one for...in loop

SanyaBratashchuk (changes requested):

agniya-i (approved):

PR #2431, opened 2023-06-27
add solution

denis-kononenko:


src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
...
+  const uniqueRobotValues = new Set(robotValues);
+  const inversedRobot = {};
+
+  if (robotValues.length !== uniqueRobotValues.size) {
+    return null;
+  }
+
+  for (const key in robot) {
+    inversedRobot[robot[key]] = key;
+  }
```
IvanFesenko, line 21: You do it with out additional operations
```suggestion
  const inversedRobot = {};

  for (const key in robot) {
    const newKey= robot[key];
    
    if (inversedRobot[newKey] !== undefined) {
      return null;
    }
    
    inversedRobot[newKey] = key;
  }
```

IvanFesenko (changes requested):
Small improvement

src/inverseRobot.js
```diff
@@ -8,6 +8,19 @@
 
 function inverseRobot(robot) {
   // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
+
+    if (inversedRobot[newKey]) {
```
volodymyr-soltys97, line 16: The condition for checking if a newKey already exists in inversedRobot is not accurate. It will fail if the newKey is '0' or any other falsy value. It is recommended to use 'hasOwnProperty' method to check if a property exists in an object.

volodymyr-soltys97 (changes requested):
Good job 👍 

volodymyr-soltys97 (approved):
Nice work 🚀

PR #2434, opened 2023-06-28
invers robot v1

oltarasov:


src/inverseRobot.js
```diff
@@ -5,9 +5,17 @@
  *
  * @return {object}
 */
-
 function inverseRobot(robot) {
-  // write code here
+  const newobj = {};
+
+  for (const prop in robot) {
+    if (newobj[robot[prop]] !== undefined) {
```
SerhiiKirik, line 12: would be better to use `hasOwnProperty` here
Also, would be better to move `robot[prop]` to the variable

SerhiiKirik (changes requested):
Well done 🔥 
Let's improve your solution 🧑‍💻

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2435, opened 2023-06-28
the inverse-robot is done

Viktoria-Moroz:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const changeProperties = {};
+
+  for (const key in robot) {
+    if (changeProperties.hasOwnProperty(key)
+    || changeProperties.hasOwnProperty(robot[key])) {
```
SerhiiKirik, line 14: you don;t need this check
```suggestion
    if (changeProperties.hasOwnProperty(robot[key])) {
```

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+      return null;
+    }
+
+    if (!changeProperties.hasOwnProperty(key)
+    || changeProperties.hasOwnProperty(robot[key])) {
+      changeProperties[robot[key]] = key;
+    }
```
SerhiiKirik, line 21: So, you can delete these checks, bc you do it before
```suggestion
      changeProperties[robot[key]] = key;
```

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+  }
+
+  const correctOrder = Object.assign({}, changeProperties);
```
SerhiiKirik, line 24: it;s unnecessary, just return `changeProperties`

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

sTorba24 (approved):
Good work!

PR #2436, opened 2023-06-29
add task solution

NikitaOrliak:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
-  // write code here
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (robot[value]) {
+      return null;
+    }
+
+    robot[value] = key;
+    delete robot[key];
```
TarasHoliuk, line 18: instead of mutating the original object, it's better to create a new one, set proper keys and values and return it
image

TarasHoliuk (changes requested):

volodymyr-soltys97 (approved):
Awesome 🚀

PR #2444, opened 2023-07-03
inverse robot

Orchibald:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const swappedObj = {};
+  const seenValues = new Set();
```
mykhalenych, line 11: The 'seenValues' Set is not necessary. You can check the existence of the property directly in the 'swappedObj'.

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (seenValues.has(value)) {
```
mykhalenych, line 16: Use 'swappedObj.hasOwnProperty(value)' instead of 'seenValues.has(value)' for checking repeated values.

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+      return null;
+    }
+    swappedObj[value] = key;
+    seenValues.add(value);
```
mykhalenych, line 20: There is no need to add the value to the 'seenValues' Set. Remove this line.

mykhalenych (changes requested):

Viktor-Kost (approved):

PR #2446, opened 2023-07-03
Develop

Roman-Bilichenko:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
Viktor-Kost, line 13: Redundant use of 'hasOwnProperty'. In this case, the for-in loop already iterates over own properties of the object, so there's no need to check if the property belongs to the object again.

Viktor-Kost (changes requested):
Nice!
Check one comment to improve your code

anastasiiavorobiova (approved):
Awesome! &#128640;

PR #2450, opened 2023-07-04
solution

nenatre:


qoa11a (changes requested):
The solution is overcomplicated.
You could create a new object, and write properties of this object as values from the argument, if there's already existing property then it means you've already added it and you should return null immediately. Otherwise, you can return this object at the end.

polosanya (approved):
Well done!

PR #2455, opened 2023-07-06
task is done

Victoriia-Melnyk:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const first = Object.keys(robot);
+  const second = Object.values(robot);
+  const newObj = {};
+
+  for (let i = 0; i < second.length; i++) {
+    for (let j = i + 1; j < second.length; j++) {
```
Anastasiia-Svintsova, line 15: Simplify the code by using `for ... in` loop
```suggestion
  const newObj = {};
  
  for (const ... in ...) { 
```

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+      if (second[i] === second[j]) {
```
Anastasiia-Svintsova, line 16: image

OleziO (approved):

PR #2462, opened 2023-07-11
add task solution

Vayts:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  if (!robot) {
+    return null;
+  }
+
+  const formatted = {};
```
roman-mirzoian, line 14: When naming, it is better to give more informative names, for example `formattedRobot` or `invertedRobot`.

roman-mirzoian (changes requested):
Good work, but it needs to be improved a little 👍

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  if (!robot) {
```
FrontKid, line 10: If you want to check if the robot is an object there is a better way to do that
```
typeof robot !== 'object' || robot === null
```
Becuz if the robot is a string or a number value you still has false in your condition
[link](https://ru.hexlet.io/qna/javascript/questions/kak-proverit-yavlyaetsya-li-peremennaya-ob-ektom-js)

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+    return null;
+  }
+
+  const formattedRobot = {};
+
+  for (const key in robot) {
+    const newKey = robot[key];
```
FrontKid, line 17: 1) [VARIABLES] - use variables for the main values so that you'll be able to reuse them and give them descriptive names.
For example, you can name them
```
    const INCOMING_ROBOT_VALUE = robot[key];
    const INCOMING_ROBOT_KEY = key;
```

FrontKid (changes requested):

src/inverseRobot.js
Vayts: It's just an argument pass check=/

src/inverseRobot.js
Vayts: On line 16, there is already a declaration of the key variable. On line 17 is a declaration of the value variable=/ I don't think upper case for variables is needed here. https://www.freecodecamp.org/news/when-to-capitalize-your-javascript-constants-4fabc0a4a4c4/

src/inverseRobot.js
FrontKid: I didn't understand what did we check here ?

dffUqp (approved):
LGTM🔥

PR #2473, opened 2023-07-15
Solution

oksana-vas:


oksana-vas:
I've found much easier solution :)

OleziO (approved):

PR #2481, opened 2023-07-18
solution

Sancz0pansa:


Matylda-Lewandowska (approved):
Your code is well readable!


In checklist in 2nd point, it is written that we shouldn't use something like that, here:
``` if (robot[key] === undefined) { ``` or
```if (robot[key]) {``` 
We should use sth like that:
```if (key in robot) {``` or even better!
```if (robot.hasOwnProperty(key)) { ```

But I can see there are no easy fixes, and precisely you didn't use forbidden cases, and code works. :)

File: src/inverseRobot.js [hunk 1]
```diff
@@ -7,7 +7,25 @@
...
+
+  for (let i = 0; i < propertyValues.length; i++) {
+    current = propertyValues[i];
+
+    for (let j = i + 1; j < propertyValues.length; j++) {
+      if (current === propertyValues[j]) {
+        return null;
+      }
+    }
+  }
```
choeqq, line 22: this loop isn't necessary - it only adds time complexity

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+  const propertyValues = Object.values(robot);
+  let current;
```
choeqq, line 12: those aren't really necessary

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+    for (let j = i + 1; j < propertyValues.length; j++) {
+      if (current === propertyValues[j]) {
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    inverse[robot[key]] = key;
+  }
```
choeqq, line 26: please create variable for robot[key]
```suggestion
  for (const key in robot) {
    if(inverse has already this property) { don't add another key-value pair }
  
    inverse[robot[key]] = key;
  }
```

choeqq (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+    }
+  }
+
+  for (const key in robot) {
+    const objPropertyValue = robot[key];
+
+    if (!inverse.hasOwnProperty(objPropertyValue)) {
+      inverse[objPropertyValue] = key;
+    }
+  }
```
choeqq, line 25:
```suggestion
  for (const key in robot) {
    const objPropertyValue = robot[key];

    if (inverse.hasOwnProperty(objPropertyValue)) {
      return null;
    }
    
    inverse[objPropertyValue] = key;
  }
```
like that there is no need to do another loop and another if statement, we can also omit `popertyValues`

choeqq (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+  const propertyValues = Object.values(robot);
+
+  for (let i = 0; i < propertyValues.length; i++) {
+    if (propertyValues.includes(propertyValues[i], i + 1)) {
+      return null;
+    }
+  }
```
asakevych, line 17: it's not necessary, you do the same in 22-24 lines

asakevych (changes requested):
Good job! you just missed one small thing 😉

choeqq (changes requested):
Almost there, waiting for any updates @Sancz0pansa 👀

choeqq (approved):
Good job ✅

PR #2483, opened 2023-07-18
task solution

ZakZh:


src/inverseRobot.js
```diff
@@ -4,10 +4,20 @@
  * @param {object} robot
  *
```
OleziO, line 5: Remove all comments from the code.

OleziO (changes requested):

IvanFesenko (approved):

PR #2484, opened 2023-07-19
added solution

FurArt:
Hi, added solution, but don't understand how to use:
GOOD EXAMPLE:
 if (key in robot) {}

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -2,11 +2,28 @@
 
 /*
  * @param {object} robot
- *
- * @return {object}
+*
+* @return {object}
 */
 
 function inverseRobot(robot) {
+  const newRobot = {};
+  let checkDoubleProper = false;
+
+  Object.entries(robot).forEach((key) => {
+    if (newRobot.hasOwnProperty(key[1])) {
+      checkDoubleProper = null;
+
+      return;
+    }
+    newRobot[key[1]] = key[0];
+  });
+
+  if (checkDoubleProper === null) {
+    return null;
+  }
+
+  return newRobot;
   // write code here
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7563 @@
 {
   "name": "js-task-boilerplate",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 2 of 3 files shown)


src/inverseRobot.js
```diff
@@ -2,11 +2,28 @@
...
+    }
+    newRobot[key[1]] = key[0];
+  });
+
+  if (checkDoubleProper === null) {
+    return null;
+  }
+
+  return newRobot;
   // write code here
```
OleziO, line 27: Remove comment from the code.

OleziO (changes requested):

src/inverseRobot.js
```diff
@@ -2,12 +2,28 @@
...
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  let checkDoubleProper = false;
+
+  Object.entries(robot).forEach((key) => {
+    if (newRobot.hasOwnProperty(key[1])) {
+      checkDoubleProper = null;
```
volodymyr-soltys97, line 15: Instead of setting 'checkDoubleProper' to null and then returning, you should return null directly. This would stop the execution of the function and would make the code cleaner and more efficient.

src/inverseRobot.js
```diff
@@ -2,12 +2,28 @@
...
+
+      return;
+    }
+    newRobot[key[1]] = key[0];
+  });
+
+  if (checkDoubleProper === null) {
+    return null;
```
volodymyr-soltys97, line 23: The condition 'if (checkDoubleProper === null)' is unnecessary because if the code execution reaches this point, it means that there are no duplicate properties and you can directly return the 'newRobot' object.

src/inverseRobot.js
```diff
@@ -2,12 +2,28 @@
...
- *
- * @return {object}
+*
+* @return {object}
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+  let checkDoubleProper = false;
```
volodymyr-soltys97, line 11: The variable 'checkDoubleProper' is misleading and not needed. Its purpose is to check if there are duplicate properties, but this can be achieved without this variable.

volodymyr-soltys97 (changes requested):
Good job 👍 

volodymyr-soltys97 (approved):

PR #2485, opened 2023-07-19
Solution

HaniaNassalska:


src/inverseRobot.js
mvjl000, line 1: Good job with the task 🚀 
However I think we can tune it even more!
Instead of standard for loop try to use [for in loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in). With this, you will be able to solve this without `properRobotKeys` and `properRobotValues` constants.

src/inverseRobot.js
```diff
@@ -7,7 +7,25 @@
...
+
+    if (properRobot.hasOwnProperty(robotKey)) {
+      return null;
+    }
+
+    properRobot = {
+      ...properRobot,
+      [robotKey]: robotValue,
+    };
+  }
```
mvjl000, line 26: Honestly I would declare the properRobot with const and simply assign new property this way (naming is up to you obviously):
 ```suggestion
    properRobot[someKey] = someValue;
```

mvjl000 (changes requested):

JoyCoffeeAddict (approved):
Great job :fire:

PR #2486, opened 2023-07-19
Solution

szymonpachucki:


src/inverseRobot.js
```diff
@@ -3,11 +3,25 @@
 /*
  * @param {object} robot
  *
- * @return {object}
+ * @return {object | null}
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixed = {};
```
mvjl000, line 10: I would change the name of this constant so it's more descriptive. Maybe `fixedRobot`?

src/inverseRobot.js
```diff
@@ -3,11 +3,25 @@
...
+  const valuesSet = new Set();
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (valuesSet.has(value)) {
```
mvjl000, line 16: I really appreciate that you are going forward and that you are using such things as Set.
But I think that in this case it's not the way. You are storing all properties in two places: fixed and valuesSet.
You could do it only using regular object and as shown in the checklist - use hasOwnProperty.
But if you really want to use some more advanced JS features, I would get rid of fixedRobot object and use Map. At the end of the function convert the map to an object and return it.

mvjl000: So if you want more complex solution - google Map and converting it to an object ;)

mvjl000 (changes requested):

src/inverseRobot.js
```diff
@@ -1,13 +1,31 @@
...
  *
- * @return {object}
-*/
+ * @return {object | null}
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobotMap = new Map();
+  const valuesSet = new Set();
```
mvjl000, line 11: You don't need a Set. Do it only with Map ;)

mvjl000 (changes requested):

choeqq (changes requested):
Waiting for any updates @szymonpachucki 👀

src/inverseRobot.js
```diff
@@ -1,13 +1,31 @@
...
+ * @return {object | null}
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobotMap = new Map();
+  const valuesArray = [];
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
asakevych, line 14: it has no sense, key would definitely be in robot, because you iterate through robot😉

src/inverseRobot.js
```diff
@@ -1,13 +1,31 @@
...
  *
- * @return {object}
-*/
+ * @return {object | null}
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobotMap = new Map();
+  const valuesArray = [];
```
asakevych, line 11: no need to use array here

src/inverseRobot.js
```diff
@@ -1,13 +1,31 @@
...
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
+      const value = robot[key];
+
+      if (valuesArray.includes(value)) {
+        return null;
+      }
```
asakevych, line 19: you can simplify it - `Map.has()` -> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/has

asakevych (changes requested):
Good job! just some changes are needed

- I appriciate you for trying to use more complex solutions, but in programming it's a good practice to use the easiest solution
- there is something like algrithms complexity and we always want our algorithms to work faster

when you are trying to do this task using arrays and  maps and their methods, you need to iterate throught every element to find the key that you need, while when using object you just know the name of key and place where you need to do your operations

i think this solution is ok, but please remember about complexity next time 😉

choeqq (changes requested):
Waiting for any updates @szymonpachucki 👀

src/inverseRobot.js
JoyCoffeeAddict: I agree, you want to check `invertedRobot.hasOwnProperty()` to check for duplicates. 
`robot.hasOwnProperty(key)`  always resolves to true.

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):
I will approve it, but I believe that you could find a solution without `invertedValues` .

PR #2487, opened 2023-07-19
Develop

spojrzenie:


src/inverseRobot.js
```diff
@@ -7,7 +7,28 @@
 */
 
 function inverseRobot(robot) {
```
darokrk, line 9: Hmm this code looks a bit complicated, you don't need two `for` loops to get object values and compare them, also to check if an object value appeared earlier and return null you should use the `hasOwnProperty()` method. Here is a simpler version:

```suggestion
function inverseRobot(robot) {
  const robotInverse = {};

  for (const key in robot) {
    if (robotInverse.hasOwnProperty(robot[key])) {
      return null;
    }

    robotInverse[robot[key]] = key;
  }

  return robotInverse;
}
```

Easier and cleaner to read 😃

darokrk (changes requested):
Please improve code with mention to be simpler and more readable 👍🏽

src/inverseRobot.js
spojrzenie: Thank you for suggestions :)

DorotaLeniecDev (approved):
Good job 💪

PR #2490, opened 2023-07-20
Solution

NNeKiTT:


src/inverseRobot.js
```diff
@@ -8,6 +8,27 @@
 
 function inverseRobot(robot) {
   // write code here
+  let tempKyesRobot = {};
```
volodymyr-soltys97, line 11: You don't need to declare an empty object before assigning a value to 'tempKyesRobot' and 'tempValueRobot'. You can directly assign the values.

src/inverseRobot.js
```diff
@@ -8,6 +8,27 @@
...
+      return null;
+    }
+  }
+
+  for (const key in robot) {
+    delete robot[key];
+  }
+
+  for (const i in tempKyesRobot) {
+    robot[tempValueRobot[i]] = tempKyesRobot[i];
```
volodymyr-soltys97, line 28: You're using 'tempValueRobot' as an index to 'tempKyesRobot', but earlier you derived these arrays from the same object. This might create an incorrect mapping if the values are not unique or not in the correct order. Instead, you should iterate the original robot object and swap the key-value pairs directly.

src/inverseRobot.js
```diff
@@ -8,6 +8,27 @@
 
 function inverseRobot(robot) {
   // write code here
+  let tempKyesRobot = {};
+  let tempValueRobot = {};
+
+  tempKyesRobot = Object.keys(robot);
+  tempValueRobot = Object.values(robot);
```
volodymyr-soltys97, line 15: You don't need to create a variable for the keys and values. You can iterate through the object directly.

src/inverseRobot.js
```diff
@@ -8,6 +8,27 @@
...
+
+  for (let i = 0; i < tempValueRobot.length; i++) {
+    if (tempValueRobot.indexOf(tempValueRobot[i]) !== i) {
+      return null;
+    }
+  }
+
+  for (const key in robot) {
```
volodymyr-soltys97, line 23: Deleting all the keys in the robot object is not necessary and can lead to unexpected behavior.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/inverseRobot.js
NNeKiTT: Тобто тимчасові змінні не потрібні? переробити без них?

PR #2492, opened 2023-07-20
Solution

in7264:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
-}
+  const InvRobot = {};
```
Andriy-Shymkiv, line 10: please use camelCase
```suggestion
  const inversedRobot = {};
```

Andriy-Shymkiv (changes requested):
good job! left comments

Andriy-Shymkiv (approved):

PR #2494, opened 2023-07-20
Created solution

vadymboichenkoAcupower:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const ReversedRobot = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
volodymyr-soltys97, line 13: The check for 'robot.hasOwnProperty(key)' is not necessary in this context as 'for...in' loop goes over the enumerable properties of an object. You can safely remove this statement.

volodymyr-soltys97 (changes requested):
Good job 👍 

nazarmatsevych (approved):

PR #2503, opened 2023-07-23
solution

cardinal1991:


src/inverseRobot.js
```diff
@@ -8,6 +8,23 @@
 
 function inverseRobot(robot) {
   // write code here
+
+  const reversed = {};
+  const check = {};
```
choeqq, line 13: we don't really need `check` object - after each loop iteration values will be updated and then we can check if property already exists

src/inverseRobot.js
```diff
@@ -8,6 +8,23 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (check[value] !== undefined) {
```
choeqq, line 18: you can use robot object for that, please check [checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) point 2

src/inverseRobot.js
```diff
@@ -8,6 +8,23 @@
...
+      return null;
+    }
+
+    reversed[value] = key;
+
+    check[value] = key;
```
choeqq, line 24: that's the same operation - we can omit that

choeqq (changes requested):

choeqq (approved):
Good job ✅

PR #2508, opened 2023-07-24
Solution

PawelP2001:


asakevych (changes requested):
Good job! small change is needed

PR #2509, opened 2023-07-24
Solution

klaudia-klis:


HaniaNassalska (approved):
It looks great! 🔥 🙂

src/inverseRobot.js
```diff
@@ -7,7 +7,26 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  function checkForRepetition(object) {
+    const objectValues = Object.values(object);
+    const setOfValues = new Set(objectValues);
+
+    return Array.from(setOfValues).length === objectValues.length;
+  }
```
choeqq, line 15: that's smart but I think for this particular case it is overkill - we can check if object value is repeating using `Object.hasOwnProperty()` method, then this function won't be necessary

choeqq (changes requested):

choeqq (approved):
Clean solution, nicely done! ✅

PR #2511, opened 2023-07-24
inverseRobot task solution

rmpolonskiy:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
-  // write code here
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    reversedRobot[value] = key;
+  }
+
+  if (Object.values(reversedRobot).length < Object.values(robot).length) {
```
TarasHoliuk, line 18: it works, but let's improve it a bit

it's better to check `If any of the object values are repeated` inside the loop
It will look clearer

TarasHoliuk (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reversedRobot = {};
+  const valueCount = {};
```
Viktor-Kost, line 11: this variable is redundant
```suggestion
```

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+
+    reversedRobot[value] = key;
+    valueCount[value] = (valueCount[value] || 0) + 1;
+  }
+
+  for (const count of Object.values(valueCount)) {
+    if (count > 1) {
+      return null;
+    }
+  }
```
Viktor-Kost, line 24: let's improve a little
1. no needs to invoke a loop twice
2. you can just use only one check
```suggestion
  for (const key in robot) {
    if (reversedRobot.hasOwnProperty(robot[key])) {
      return null;
    }

    reversedRobot[robot[key]] = key;
  }
```

Viktor-Kost (changes requested):

src/inverseRobot.js
rmpolonskiy: Oh, I got it 
Thank you!

etojeDenys (approved):
lgtm

PR #2513, opened 2023-07-24
Solution

jakubMikolajczykTech:


src/inverseRobot.js
```diff
@@ -1,13 +1,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    newRobot[robot[key]] = String(key);
```
choeqq, line 11: key is always string - objects can't have anything other than that - [Maps ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)can 😄

src/inverseRobot.js
```diff
@@ -1,13 +1,17 @@
...
- *
- * @return {object}
-*/
-
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
```
choeqq, line 7: please create variable for `robot[key]` - you're using it twice

choeqq (changes requested):

choeqq (changes requested):
Waiting for any updates @kabakTech 👀

JoyCoffeeAddict (approved):

PR #2516, opened 2023-07-25
add task solution

Nightycake:


src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
-  // write code here
+  const inverseObj = {};
+  const repeatedVal = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (repeatedVal[value]) {
+      return null;
+    }
```
choeqq, line 18: you can check if there is repeatedVal by checking if `inverseObj` has this property already - you don't need to use another object to store that

choeqq (changes requested):

choeqq (approved):
Nicely done 🚀

PR #2517, opened 2023-07-25
solution

MaciejKujawa:


src/inverseRobot.js
```diff
@@ -4,10 +4,31 @@
  * @param {object} robot
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInversed = {};
+  let robotValues = [];
```
choeqq, line 11: robot values are stored in `robotInversed[key]` - we don't need another object

src/inverseRobot.js
```diff
@@ -4,10 +4,31 @@
...
+    for (let j = 0; j < robotValues.length; j++) {
+      if (i === j) {
+        continue;
+      }
+
+      if (robotValues[i] === robotValues[j]) {
+        return null;
+      }
+    }
+  }
```
choeqq, line 25: this step isn't necessary - we can do all that in other loop, down below

src/inverseRobot.js
```diff
@@ -4,10 +4,31 @@
...
+
+  for (const key in robot) {
+    robotInversed[robot[key]] = key;
+  }
```
choeqq, line 29: please create variable for `robot[key]`
```suggestion
  for (const key in robot) {
    if(check if property already exists) { if exists do something }
    other than that - do something else
    robotInversed[robot[key]] = key;
  }
```

choeqq (changes requested):

src/inverseRobot.js
```diff
@@ -4,10 +4,22 @@
...
+  const robotInversed = {};
+
+  for (const key in robot) {
+    const a = robot[key];
+
+    if (robotInversed.hasOwnProperty(a)) {
+      return null;
+    } else {
+      robotInversed[a] = key;
+    }
```
choeqq, line 19: if statement returns or passes on, so the code after will always be executed or exited out of function - there is no need to use `else`

choeqq (approved):

PR #2521, opened 2023-07-25
solution

olegobiukh:


src/inverseRobot.js
```diff
@@ -4,10 +4,26 @@
...
+  }
+
+  const robotLength = Object.keys(robot).length;
+  const objLength = Object.keys(obj).length;
+
+  if (robotLength === objLength) {
+    return obj;
+  }
+
+  return null;
```
DarkMistyRoom, line 26: better to check it on each iteration to return null on the first repeated value and reduce the algorithm complexity

DarkMistyRoom (changes requested):

etojeDenys (approved):
lgtm

PR #2522, opened 2023-07-25
add task solution

Reject-4444:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
```
xiimiik, line 10: Avoid using unnecessary operations. Here, you are creating array of values but use it only once. You can directly check the robot object in the for loop

xiimiik (changes requested):
GJ, left few comments

Viktor-Kost (approved):

PR #2523, opened 2023-07-25
add task solution

MateuszSlezak:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const correctRobot = {};
+
+  for (const key in robot) {
+    const KEY_IS_DUPLICATED = robot[key] in correctRobot;
```
choeqq, line 13: typically we use uppercased constants for harcoded values for example: days of week or specific number, please change that to camelCase

choeqq (changes requested):

choeqq (approved):
Good job 🚀

PR #2524, opened 2023-07-25
Solution

SergiyVS:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot.hasOwnProperty(robot[key])) {
+      return null;
+    };
+
+    newRobot[robot[key]] = key;
```
maxim2310, line 17: use variable for robot[key]

maxim2310 (changes requested):
Good Job!!!
just please fix one mistakes

src/inverseRobot.js
SergiyVS: Done

maxim2310 (approved):
Good job!!!

PR #2526, opened 2023-07-26
adding a solution

vensecue:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverseObj = {};
+  const valueKeys = [];
```
choeqq, line 11: we don't need to use separate array for storing `valueKeys` we can just check if the value already exists in our object - if so we don't want to add it

choeqq (changes requested):

choeqq (approved):
Good job! 🚀

PR #2527, opened 2023-07-26
add inverse solution

ghost:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
-  // write code here
+  const robotKeys = Object.keys(robot);
+  const newRobot = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
+    if (newRobot.hasOwnProperty(robot[robotKeys[i]])) {
+      return null;
+    }
+    newRobot[robot[robotKeys[i]]] = robotKeys[i];
+  }
```
choeqq, line 18: Please use variable for `robotKeys[i]`

choeqq (changes requested):

choeqq (changes requested):
Waiting for any updates @WaveOfDandelions 👀

src/inverseRobot.js
WaveOfDandelions: > Please use variable for `robotKeys[i]`

Can you show me example what you exactly mean?
 I have variable before in code const robotKeys = Object.keys(robot);

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotKeys = Object.keys(robot);
+  const newRobot = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
+    if (newRobot.hasOwnProperty(robot[robotKeys[i]])) {
```
choeqq, line 14:
```suggestion
  for (let i = 0; i < robotKeys.length; i++) {
   const robotKey = robotKeys[i];
    if (newRobot.hasOwnProperty(robot[robotKey])) {
```

choeqq (changes requested):
Is that fine?

src/inverseRobot.js
asakevych: 1. Kacper means you need to create variable with `robotKeys[i]` and use it inside loop, so your code would be easier to read
2. I think you can use `for in` loop through object insted of creating array with robot keys to simplify your solution

asakevych (changes requested):

WaveOfDandelions:
> Is that fine?

Ok I will rewrite

choeqq (approved):
Good job ✅

PR #2534, opened 2023-07-27
feat: the solution of the task (Sun, Jul 09, 2023 06:24:49 PM)

FrontKid:


AndriiKuz (approved):
Good job! it looks nice)

src/inverseRobot.js
```diff
@@ -7,7 +7,33 @@
...
+  // check if the robot is an object
+  if (typeof robot !== 'object' || robot === null) {
+    return null;
+  }
+
+  // "robot"'s keys become values and values become "robot"'s keys
+  const turnedRobot = {};
+
+  for (const key in robot) {
+    const INCOMING_ROBOT_VALUE = robot[key];
```
dffUqp, line 19: it's better to rename these variables to lowercase, basically we use upper + snake keys for static variables that won't be changed

learn more about javascript naming convention - [link](https://www.robinwieruch.de/javascript-naming-conventions/)

src/inverseRobot.js
```diff
@@ -7,7 +7,33 @@
...
+    const INCOMING_ROBOT_KEY = key;
+
+    // Check if there are already such values in the created object
+    const hasDoubleValue
+      = {}.prototype.hasOwnProperty.call(turnedRobot, INCOMING_ROBOT_VALUE);
```
dffUqp, line 24: it looks like you wanted to use `__proto__` here instead of `prototype`.

I think this solution is creative but a little bit :exploding_head: let's use `Object.hasOwnProperty` it makes our code shorter and more readable

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

src/inverseRobot.js
FrontKid: But why linter tell us to use Object.prototype... instead  hasOwnProperty?

roman-mirzoian (approved):
Well done 👍

src/inverseRobot.js
dffUqp: Here is a [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) where you can read more about the object, pay attention to this text:

> A method can be called on most JavaScript objects because most objects derive from Object and therefore inherit its methods.

By writing `Object.prototype`, we know 100% that we have a `hasOwnProperty` method, and we will get some results, not an error.



Some useful links:
[stackoverflow thread about this eslint rule](https://stackoverflow.com/questions/39282873/object-hasownproperty-yields-the-eslint-no-prototype-builtins-error-how-to)
[proto-vs-prototype](https://javascript.plainenglish.io/proto-vs-prototype-in-js-140b9b9c8cd5)


You were trying to get a prototype for an empty object that doesn't have it, so I wrote you that comment to ask you to make your code work, sorry if it was not clear.

PR #2535, opened 2023-07-27
Develop

AndriiKuz:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (inversedRobot.hasOwnProperty(robot[key])) {
```
FrontKid, line 13: it's better to use Object.prototype.hasOwnProperty.call(inversedRobot, robot[key])
[link](https://eslint.org/docs/latest/rules/no-prototype-builtins)

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+      return null;
+    }
+    inversedRobot[robot[key]] = key;
```
FrontKid, line 16: And you can give variables more meaningful names
For example:
 ```
const incomingRobotValue = robot[key]
```

**Review by FrontKid — CHANGES_REQUESTED** [review]

File: src/inverseRobot.js (hunk 1 above)
Comment (AndriiKuz) [reply src/inverseRobot.js]: Maybe you're right, but I don't know how to work with prototypes yet

File: src/inverseRobot.js [hunk 3]
```diff
@@ -7,7 +7,18 @@
...
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const incomingRobotValue = robot[key];
+
+    if (inversedRobot.hasOwnProperty(incomingRobotValue)) {
+      return null;
+    }
+    inversedRobot[incomingRobotValue] = key;
```
dffUqp, line 18:
```suggestion
    if (inversedRobot.hasOwnProperty(incomingRobotValue)) {
      return null;
    }
    
    inversedRobot[incomingRobotValue] = key;
```

dffUqp (approved):
GJ✔️

PR #2536, opened 2023-07-27
solved task

bormaxvel:


src/inverseRobot.js
```diff
@@ -7,7 +7,22 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertRobot = {};
+  let prewPseudoVallue = '';
```
loralevitska, line 11: No need use `pseudo` and you should fix typo in word `value` and `prev` (shortand from previous word)
```suggestion
  let prevValue = '';
```

loralevitska (approved):
Well done)

PR #2543, opened 2023-07-28
Solution

polinavafik:


src/inverseRobot.js
```diff
@@ -6,8 +6,45 @@
...
+      // we are getting the actual key name
+      // and we store it in variable, named 'value'
+
+      if (!fixedRobot.hasOwnProperty(value)) {
+        fixedRobot[value] = key;
+      } else {
+        return null;
+      }
+    }
+  }
```
dffUqp, line 45: I think we can avoid using nested  `if`  here 
try to check only robot value

dffUqp (changes requested):
Almost done:smiley:
just one moment to fix

src/inverseRobot.js
```diff
@@ -6,8 +6,39 @@
...
+ inverseRobot(kolli) === { name: 'Kolli', chipVer: '123', wheels: '3' }
+```
+*/
+
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const key in robot) {
+    if (fixedRobot.hasOwnProperty(robot[key])) {
```
Moroz-Dmytro, line 34: We can create variable for `robot[key]`, we use it 2 times

Moroz-Dmytro (approved):
👍

PR #2546, opened 2023-07-30
add solution

KatOlista:


Koliras (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotValues = Object.values(robot);
+  const robotKeys = Object.keys(robot);
+  const newObj = {};
+
+  for (const value of robotValues) {
```
dffUqp, line 14: let try to use for in loop to go through object

lear more about `for in`  - [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
...
+    if (robotValues.indexOf(value) !== robotValues.lastIndexOf(value)) {
+      return null;
+    }
+  }
+
+  for (let i = 0; i < robotKeys.length; i++) {
```
dffUqp, line 20: Try to rewrite it using only one loop.

1.  check if we already have this key in `newObj`
2.  add key to `newObj`

dffUqp (changes requested):
Almost done:smiley:

src/inverseRobot.js
```diff
@@ -7,7 +7,46 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const outRobot = { ...robot };
+  const newRobot = {};
+  const keysArr = [];
+  const valuesArr = [];
+
+  for (const key in outRobot) {
+    keysArr.push(key);
+  }
```
Moroz-Dmytro, line 17: You don't need this, try to rewrite it. But for the future, if you want to get an array of all keys you can use [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

src/inverseRobot.js
```diff
@@ -7,7 +7,46 @@
...
+
+  for (let i = 0; i < keysArr.length; i++) {
+    valuesArr.push(outRobot[keysArr[i]]);
+  }
```
Moroz-Dmytro, line 21: You don't need this, try to rewrite it. But for the future, if you want to get an array of all values you can use [Object.values()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values)

src/inverseRobot.js
```diff
@@ -7,7 +7,46 @@
...
+
+  for (let i = 0; i < valuesArr.length; i++) {
+    if (newRobot[valuesArr[i]]) {
+      return null;
+    }
+
+    newRobot[valuesArr[i]] = keysArr[i];
+  }
+
+  return { ...newRobot };
```
Moroz-Dmytro, line 31: Try to rewrite it using `for in` loop

src/inverseRobot.js
```diff
@@ -7,7 +7,46 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const outRobot = { ...robot };
+  const newRobot = {};
+  const keysArr = [];
+  const valuesArr = [];
```
Moroz-Dmytro, line 13: Try to make up better names for variables(but you don't need so many, try to refactor this)

Moroz-Dmytro (changes requested):
It works, but try to simplify your solution
Comment about rewriting it using 1 loop is not fixed, now we have even more loops than before fix 😁
If you have problems with that you can ask help in qna chat

src/inverseRobot.js
```diff
@@ -7,7 +7,35 @@
...
+  const newRobot = {};
+
+  for (const key in incomeRobot) {
+    if (newRobot[incomeRobot[key]]) {
+      return null;
+    }
+    newRobot[incomeRobot[key]] = key;
+  }
+
+  return { ...newRobot };
```
Moroz-Dmytro, line 20: Also don't need to make a copy

src/inverseRobot.js
```diff
@@ -7,7 +7,35 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const incomeRobot = { ...robot };
+  const newRobot = {};
+
+  for (const key in incomeRobot) {
+    if (newRobot[incomeRobot[key]]) {
+      return null;
+    }
+    newRobot[incomeRobot[key]] = key;
```
Moroz-Dmytro, line 17:
```suggestion
    }
    
    newRobot[incomeRobot[key]] = key;
```

Moroz-Dmytro (approved):
👍 
Let's go ahead with this solution 
But please, check the comments below to improve your code in future

PR #2548, opened 2023-07-30
add task solution

Koliras:


AntonHerasymov (approved):
Good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const KEYS = Object.values(robot);
+  const VALUES = Object.keys(robot);
```
dffUqp, line 11: I think we can make this shorter, let's get robot value in `for` loop and then use something like this:
image

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+  const INVERSED_ROBOT = {};
```
dffUqp, line 12: it's better to rename this variable to lowercase, basically we use upper + snake keys for static variables that won't be changed

learn more about javascript naming convention - [link](https://www.robinwieruch.de/javascript-naming-conventions/)

dffUqp (changes requested):
Almost done:smiley:

BudnikOleksii (approved):
Nicely done👍

PR #2557, opened 2023-07-31
Solution

AndrMar1939:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+  }
+
+  const arrayOfKeys = Object.keys(robot);
+  const inverseObj = {};
+
+  arrayOfKeys.forEach((key) => {
+    const newKey = robot[key];
+
+    inverseObj[newKey] = key;
+  });
```
dffUqp, line 24: let's try to use `for in` loop to go through object

learn more about `for in`  - [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

dffUqp (changes requested):
Almost done:smiley:

Read second point from [checklist](https://github.com/mate-academy/js_inverse_robot/blob/master/checklist.md) there is a good hint to make this code shorter

src/inverseRobot.js
AndrMar1939: thank you!

alexander-ignatow (approved):
Looks good :+1:

PR #2560, opened 2023-07-31
Solved task

vserhiichuk:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const REVERSE_ROBOT = {};
```
voievodik, line 10: It's better to use camelCase construction, because in the form of uppercase letters, we explicitly set for some specific value that can be used globally

voievodik (changes requested):
Good job!

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
+  const VALUES = Object.values(robot);
+  const KEYS = Object.keys(robot);
+  const REMOVE_DUBLICATE_VALUES = new Set(VALUES);
+
+  if ([...REMOVE_DUBLICATE_VALUES].length !== VALUES.length) {
```
Wesses, line 15: I think it would be better to use the object function to define the repeating field. Especially considering the fact that we were given a hint in the checklist. [Cheklist - paragraph 2].
`if (robot.hasOwnProperty(key))`
But I`m not sure about this, u can re push your project and mentors will say their opinion.

Wesses (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const reverseRobot = {};
+  const VALUES = Object.values(robot);
+  const KEYS = Object.keys(robot);
+  const REMOVE_DUBLICATE_VALUES = new Set(VALUES);
```
Moroz-Dmytro, line 13: Variable names should be in camelCase

Moroz-Dmytro (changes requested):
Try to simplify your solution
P.S. You can check in the loop if your `reversedRobot` already contains such key

src/inverseRobot.js
Moroz-Dmytro: Good point

dffUqp (approved):
LGTM🔥👀

PR #2561, opened 2023-07-31
Add task solution

davinnchii:


src/inverseRobot.js
```diff
@@ -8,6 +8,27 @@
...
+  const reversedRobot = {};
+
+  for (const key in robot) {
+    reversedRobot[robot[key]] = key;
+  }
+
+  // Check our input object for duplicate values
+  const isDuplicates = Object.values(robot).map(function(item, idx, arr) {
+    return arr.indexOf(item) !== idx;
+  });
```
vserhiichuk, line 20: I think it would be better to use the object function to define the repeating field. Especially considering the fact that we were given a hint in the checklist. [Cheklist - paragraph 2].
if (`robot.hasOwnProperty(key)`)

vserhiichuk (approved):
It looks great!

dffUqp (approved):
Looks good👍

PR #2562, opened 2023-07-31
Solution

voronine:


KlymSudakov (changes requested):
Rename new object name to something more descriptive.
Instead of using Object.assign, you can simply add the property directly to the new object, which will make the code clearer.
You could use a variable for a value of object's key which will make the code clearer as well.
You also could use 'hasOwnProperty()' method to check if the property key exists directly on the object robot.

src/inverseRobot.js
```diff
@@ -2,12 +2,31 @@
...
  * @param {object} robot
- *
+ *Example:
+```
+ const kolli = { Kolli: 'name', 123: 'chipVer', 3: 'wheels' };
+ const robert = { Robert: 'name', 123: 'chipVer', 113: 'chipVer' };
+ inverseRobot(robert) === null
+ inverseRobot(kolli) === { name: 'Kolli', chipVer: '123', wheels: '3' }
+```
+
```
Moroz-Dmytro, line 12: Don't leave comments in code

src/inverseRobot.js
```diff
@@ -2,12 +2,31 @@
...
  * @return {object}
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
```
Moroz-Dmytro, line 17: I'd better name it `inversedRobot`, it's more clear name

src/inverseRobot.js
```diff
@@ -2,12 +2,31 @@
...
+
+  for (const key in robot) {
+    const robotItem = robot[key];
```
Moroz-Dmytro, line 20: Also try to make up something more descriptive

Moroz-Dmytro (changes requested):
Good work 👍 
Only small recomendations

BudnikOleksii (approved):
Nicely done🚀

PR #2566, opened 2023-07-31
solution

Yevheniia-Sid:


src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotInverse = {};
+
+  for (const key in robot) {
+    const value = robot[key];
```
kaavka, line 13: Do we realy need these variable?

kaavka (approved):

src/inverseRobot.js
kaavka: Forgot to say that everythin else looks awesome. Nice job

Moroz-Dmytro: 👍 
But try to give more descriptive names, not just `value`

src/inverseRobot.js
```diff
@@ -7,7 +7,18 @@
...
+
+    if (robotInverse.hasOwnProperty(value)) {
+      return null;
+    }
+    robotInverse[value] = key;
```
Moroz-Dmytro, line 18:
```suggestion
    }
    
    robotInverse[value] = key;
```

Moroz-Dmytro (approved):
👍

PR #2568, opened 2023-07-31
task solution

no4kar:


src/inverseRobot.js
```diff
@@ -1,13 +1,20 @@
...
-*/
-
+ * @return {object | null}
+ */
 function inverseRobot(robot) {
-  // write code here
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    if (inversedRobot.hasOwnProperty(robot[key])) {
```
loralevitska, line 11: It's better to use 'Object.prototype.hasOwnProperty.call(inversedRobot, robot[key])' instead of 'inversedRobot.hasOwnProperty(robot[key])' to check if a property exists. This ensures the check works even if a property named 'hasOwnProperty' exists on the object.

src/inverseRobot.js
```diff
@@ -1,13 +1,20 @@
...
+      return null;
+    }
+    inversedRobot[robot[key]] = key;
```
loralevitska, line 14:
```suggestion
    }
    
    inversedRobot[robot[key]] = key;
```

loralevitska (approved):
Good job!

PR #2576, opened 2023-08-01
add task solution

rostyslav48:
BTW I think it would be better to add the following condition:
```
    if (!isNaN(+robotKeys[i])) {
      inversedRobotObj[robotValues[i]] = +robotKeys[i];
      continue;
    }
```

in this way,  we will get a number data where it should be, as a result it would be easier to operate with them in the future

### DIFF
--- a/src/inverseRobot.js
+++ b/src/inverseRobot.js
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotKeys = Object.keys(robot);
+  const robotValues = Object.values(robot);
+  const inversedRobotObj = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
+    if (inversedRobotObj.hasOwnProperty(robotValues[i])) {
+      return null;
+    }
+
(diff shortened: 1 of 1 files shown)


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const robotKeys = Object.keys(robot);
+  const robotValues = Object.values(robot);
+  const inversedRobotObj = {};
+
+  for (let i = 0; i < robotKeys.length; i++) {
```
annak1104, line 14: maybe, in this case, it might be better to use a loop "for in" for a cleaner and more readable code.

annak1104 (approved):
Great job!

src/inverseRobot.js
rostyslav48: Yep, I fully agree with you. Thanks a lot for your comment!

rostyslav48: I have changed my code a little bit

dffUqp (approved):
Perfect🚀

PR #2580, opened 2023-08-01
task_solution

BeginerAlex:


src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
 function inverseRobot(robot) {
-  // write code here
+  const prevValue = new Set();
+  const copyRobot = {};
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
+      const currentValue = robot[key];
+
+      if (prevValue.has(currentValue)) {
```
polosanya, line 17: Better check if copyRobot hasOwnProperty with currentValue.
prevValue is actually redundant

polosanya (changes requested):
Almost done!

TarasHoliuk (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,21 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const copyRobot = {};
+
+  for (const key in robot) {
+    if (robot[key]) {
```
TarasHoliuk, line 13: actually, the same as before:
image

this check is redundant, for example, if the value is 0 and the key is let's say `version` you will not add the property `'0': 'version'` because of this check (0 is falsy value)

TarasHoliuk (changes requested):
Approved by mistake the first time - dismissed it (not sure if task status updates after such actions), some changes are still required

OleziO (approved):

PR #2584, opened 2023-08-01
solution inverseRobot

ihorid:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+
+  for (let i = 0; i <= valueOfRobot.length; i++) {
+    if (valueOfRobot.indexOf(valueOfRobot[i], i + 1) !== -1) {
+      return null;
+    }
+  }
+
+  for (const key in robot) {
+    repairedRobot[robot[key]] = key;
+  }
```
Moroz-Dmytro, line 21: Try to simplify your solution. We can use only one loop for the solution
P.S. Take a look at checklist, we can use `hasOwnProperty` for checking if key is already present

Moroz-Dmytro (changes requested):

Moroz-Dmytro (approved):
👍

PR #2585, opened 2023-08-01
Solution

AnastasiiaHombalevska:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newRobot = {};
+
+  for (const key in robot) {
+    if (newRobot[robot[key]]) {
```
Vitalii-Petruk, line 13: 2.[checklist] use correct check for property presence in object. Some properties can be present, but still contain falsy value.
GOOD EXAMPLE:
if (key in robot) {
EVEN BETTER EXAMPLE:
BAD EXAMPLE:
if (robot[key]) {

Vitalii-Petruk (changes requested):
Great job 👍 
But fix your code, please...
(after that 🥇)

Moroz-Dmytro (approved):
👍

PR #2590, opened 2023-08-02
add task solution

edgergard:


alexander-ignatow (approved):
Valid solution. Let's move on!

Interesting idea with the Set

However - I am not sure about the effectiveness of it
Most probably you are doing an extra job for this check

A more effective solution would be making a single loop - you can check for duplicates and set the required data with it.

PR #2591, opened 2023-08-02
solution

00Mass00:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixedRobot = {};
+
+  for (const value in robot) {
```
qoa11a, line 12: I'm unsure you should use `value` here, as it's a key. I understand that it becomes a value of the new object, but the variable should be named after what it stores.

qoa11a (approved):

Moroz-Dmytro (approved):

PR #2592, opened 2023-08-02
Solution for task

thaiduk:


olegpaunov (approved):
good job

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const inversedRobot = {};
+
+  for (const key in robot) {
+    const inversedRobotValue = robot[key];
+
+    if (inversedRobot.hasOwnProperty(robot[key])) {
+      return null;
+    }
+
+    inversedRobot[inversedRobotValue] = key;
```
alexander-ignatow, line 19: You can either remove `inversedRobotValue` or use it everywhere like this:

```suggestion
    const inversedRobotValue = robot[key];

    if (inversedRobot.hasOwnProperty(inversedRobotValue) {
      return null;
    }

    inversedRobot[inversedRobotValue] = key;
```

alexander-ignatow (approved):
Good 👍 

Please see the comment below 👇

PR #2595, opened 2023-08-02
add task solution

cty2802:


k1NAs (approved):
Nicely done! :fire:

src/inverseRobot.js
```diff
@@ -4,10 +4,22 @@
  * @param {object} robot
  *
  * @return {object}
-*/
+ */
 
 function inverseRobot(robot) {
-  // write code here
+  const inversedObject = {};
```
Moroz-Dmytro, line 10: Try to make up better names for variables. Don't use the words `object`, `array` or something like that in the name. The name should describe what is saved in this variable. Here you can name it `inversedRobot` for example

Moroz-Dmytro (approved):
GJ 👍

PR #2597, opened 2023-08-02
add task solution

SmartSergiZ:


bgutai (approved):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverseObject = {};
```
Moroz-Dmytro, line 10: Try to make up a more descriptive name, for example, `inversedRobot`. Try not to use the words `object`, `array`, and so on in your variables names

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+
+  for (const key in robot) {
+    const value = robot[key];
```
Moroz-Dmytro, line 13: Also, try to make up a better name, not just `value`. Maybe now it's ok, but when your code will grow up, it will be difficult to understand what value it is

Moroz-Dmytro (changes requested):
GJ 👍 
But small comments

Anna-Tkachenko (approved):
⭐️

PR #2598, opened 2023-08-03
solution

vladyslav-svorin:


src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+      if (robot[initialKeys[i]] === robot[initialKeys[j]]) {
+        return null;
+      }
+    }
+  }
+
+  for (const key in robot) {
+    const reversedKey = robot[key];
+
+    delete robot[key];
```
loralevitska, line 23: You are modifying the input object which is generally a bad practice. It is better to create a new object for the output.

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const initialKeys = Object.keys(robot);
+
+  for (let i = 0; i < initialKeys.length; i++) {
```
loralevitska, line 12: Nested for loops for checking duplicate values are not optimal. Consider using a Set for checking duplicate values in an array as it is more efficient.

loralevitska (changes requested):
Almost done) Let's improve

VitaliyBodnar (approved):
Good job!

PR #2601, opened 2023-08-03
finish the task

serkrops:


src/inverseRobot.js
```diff
@@ -8,6 +8,24 @@
 
 function inverseRobot(robot) {
   // write code here
+
+  const inverseRobotResult = {};
```
Moroz-Dmytro, line 12: You can call it just `inversedRobot`

src/inverseRobot.js
```diff
@@ -8,6 +8,24 @@
...
+    arrayOfKeys.push(robot[key]);
+  }
+
+  arrayOfKeys.sort();
+
+  for (let i = 0; i < arrayOfKeys.length; i++) {
+    if (arrayOfKeys[i] === arrayOfKeys[i + 1]) {
+      return null;
+    }
+  }
```
Moroz-Dmytro, line 26: Try to simplify your solution and use only 1 loop.
P.S. To check if key already present in inversedRobot, you can use `hasOwnProperty`

Moroz-Dmytro (changes requested):
It works 👍 
But let's improve and simplify it

Anna-Tkachenko (approved):
⭐️

PR #2602, opened 2023-08-03
add task solution

AngerDESTROYS:


src/inverseRobot.js
```diff
@@ -8,6 +8,18 @@
...
   // write code here
+  const inverted = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (inverted[value] !== undefined) {
+      return null;
+    }
+    inverted[value] = key;
```
BudnikOleksii, line 19: [CODE STYLE] - use correct check for property presence in object. Some properties can be present, but still contain falsy value.
image

BudnikOleksii (changes requested):
Nice start👍
Let's improve the task in accordance with the checklist.

BudnikOleksii (approved):
Good job👍

PR #2603, opened 2023-08-03
add task solution

bgutai:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    if (invertedRobot.hasOwnProperty(robot[key])) {
```
Moroz-Dmytro, line 13: We can also create variable for `robot[key]` as we use this value 2 times

PR #2604, opened 2023-08-03
solution

Duster-1:


src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+    if (seenValues[value]) {
+      return null;
+    }
+    seenValues[value] = true;
+  }
+
+  const result = {};
+
+  for (const key in robot) {
+    result[robot[key]] = key;
```
nazarmatsevych, line 24: Actually second loop looks redundant, you can do the same using just one:

```
    if (seenValues.hasOwnProperty(robot[key])) {
      return null;
    }

    seenValues [robot[key]] = key;
```

nazarmatsevych (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
 
 function inverseRobot(robot) {
-  // write code here
+  const seenValues = {};
+  const result = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (seenValues[value]) {
```
TarasHoliuk, line 16: you don't need a separate `seenValues` object
if this value was seen before than the `result` object has such key

TarasHoliuk (changes requested):

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+  const result = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (result[value]) {
+      return null;
+    }
+    result[value] = true;
+    result[robot[key]] = key;
```
TarasHoliuk, line 19:
```suggestion
    result[value] = key;
```

TarasHoliuk (approved):

PR #2608, opened 2023-08-05
inverseRobot Solution

ZadorozhnyiYevhenii:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const inverse = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (Object.keys(inverse).includes(value.toString())) {
```
VitaliyBondarenko1982, line 15: redundanr to such big checking
it can be 
`if (value in inverese) {` or `if (inverse[value]) {` or `if (inverse.hasOwnProperty(value)) {`

src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
...
+      return null;
+    } else {
+      inverse[value] = key;
+    }
```
VitaliyBondarenko1982, line 19:
```suggestion
    } 
    
    inverse[value] = key;
```
try to avoid else where it is possible

VitaliyBondarenko1982 (changes requested):
Good job.
Check a couple comments

nazarmatsevych (approved):

PR #2615, opened 2023-08-07
Solution

UlianaChorna:


src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const revers = {};
+
+  for (const key in robot) {
+    if (robot[key] in revers) {
```
xiimiik: create variable for `robot[key]`, because u use it twice and its a improve readability
also `in` can work not correctly, try `.hasOwnProperty`.

src/inverseRobot.js
```diff
@@ -7,7 +7,16 @@
...
+      return null;
+    }
+    revers[ robot[key] ] = key;
```
xiimiik, line 16:
```suggestion
    revers[robot[key]] = key;
```

xiimiik (changes requested):

Esceype (approved):
Well done!!!

PR #2617, opened 2023-08-07
solution

mykyta01:


src/inverseRobot.js
```diff
@@ -8,6 +8,23 @@
 
 function inverseRobot(robot) {
   // write code here
+  const invertedRobot = {};
+  const valuesSeen = {};
```
lerastarynets, line 12: The 'valuesSeen' object is redundant. You only need to check if the 'value' already exists as a key in the 'invertedRobot' object. This will simplify the code and reduce memory usage.

src/inverseRobot.js
```diff
@@ -8,6 +8,23 @@
...
+
+  for (const key in robot) {
+    if (robot.hasOwnProperty(key)) {
```
lerastarynets, line 15: The 'hasOwnProperty' check is unnecessary in this for..in loop as we're not dealing with inherited properties in this context. It can be safely removed to cleanup the code.

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #2619, opened 2023-08-07
Solution

oirevlis1:


src/inverseRobot.js
```diff
@@ -7,7 +7,20 @@
...
+  }
+
+  const robotLength = Object.entries(robot).length;
+  const inversedRobotLength = Object.entries(inversedRobot).length;
+
+  if (robotLength !== inversedRobotLength) {
+    return null;
+  }
+
+  return inversedRobot;
```
IvanFesenko, line 23: if should check if new keu exists in loop
```suggestion
  for (const key in robot) {
    const newKey = robot[key];
    
    if (inversedRobot[newKey]) {
      return null;
    }
    
    inversedRobot[newKey] = key;
  }

  return inversedRobot;
```

IvanFesenko (changes requested):
Check my comment

OleziO (approved):

PR #2620, opened 2023-08-09
InverseRobot. Tests passed

momos1703:


src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const correctedRobot = {};
+  const robotValues = Object.values(robot);
+
+  // Check if values are not repeated
+  for (const value of robotValues) {
```
mykhalenych, line 14: you can avoid this loop

src/inverseRobot.js
```diff
@@ -7,7 +7,24 @@
...
+    if (robotValues.indexOf(value) !== robotValues.lastIndexOf(value)) {
+      return null;
+    }
+  }
+
+  // Change keys and values on places
+  for (const key in robot) {
+    const robotValue = robot[key];
+
```
mykhalenych, line 23: and add check if !correctedRobot[robotValue]

mykhalenych (changes requested):

src/inverseRobot.js
momos1703: Done!

src/inverseRobot.js
momos1703: Added

denys-danyliuk (approved):

PR #2621, opened 2023-08-10
solution was added

Liliya-Kalinichenko:


maxim2310 (approved):
Cleare and beautiful, good job

PR #2624, opened 2023-08-12
add task solution

Chumakv:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const newObject = {};
+
+  for (const [key, value] of Object.entries(robot)) {
```
volodymyr-soltys97, line 12: Using 'Object.entries()' in a for loop could cause the function to slow down, especially if the input object is large. Consider using 'Object.keys()' or 'Object.values()' to optimize performance.

src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
+    if (newObject[value] === undefined) {
```
volodymyr-soltys97, line 13: Instead of checking for undefined with 'newObject[value] === undefined', consider using 'newObject.hasOwnProperty(value)' which is a more reliable way to check if a property exists in an object.

src/inverseRobot.js
volodymyr-soltys97: Or you can use loop `for in`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #2627, opened 2023-08-13
first attempt

ColumbusDeveloper:


volodymyr-soltys97 (changes requested):
Try to simplify your solution a bit, it looks a bit complicated now, you can use a loop to work with the objects and then check if the property exists in the object

src/inverseRobot.js
```diff
@@ -7,7 +7,23 @@
...
+    NORMALIZED_ROBOT_OBJECT[RIGHT_WAY_ROBOT_KEY] = RIGHT_WAY_ROBOT_VALUE;
+
+    if (!CONTROL_IF_KEY_DUBBED.includes(RIGHT_WAY_ROBOT_KEY)) {
+      CONTROL_IF_KEY_DUBBED.push(RIGHT_WAY_ROBOT_KEY);
+    } else {
+      return null;
+    }
+  }
+
+  return NORMALIZED_ROBOT_OBJECT;
```
volodymyr-soltys97, line 26: You can rewrite like this, consider it will be better and easier
```suggestion
  const inversedRobot = {};

  for (const key in robot) {
    const inversedKey = robot[key];

    if (inversedRobot.hasOwnProperty(inversedKey)) {
      return null;
    }
    inversedRobot[inversedKey] = key;
  }

  return inversedRobot;
```

volodymyr-soltys97 (approved):

PR #2629, opened 2023-08-15
Inverse robot need to check

tetyana-shatrova:


src/inverseRobot.js
```diff
@@ -7,7 +7,17 @@
...
-  // write code here
+  const newRobot = {};
+
+  for (const [key, value] of Object.entries(robot)) {
+    if (newRobot[value]) {
+      return null;
+    }
+
+    newRobot[value] = key;
+  }
```
VitaliyBodnar, line 18: using `for in` loop will be more performant
so, please replace `for of` to `for in`

VitaliyBodnar (changes requested):
Almost done!

src/inverseRobot.js
tetyana-shatrova: hello! unfortunately it doesn't work with for in loop

VitaliyBodnar (approved):
Good job!

VitaliyBodnar (approved):

PR #2640, opened 2023-08-26
task solution

TenetaJ:


src/inverseRobot.js
```diff
@@ -7,7 +7,32 @@
...
+    const lenValue = equalValue[value].length;
+
+    if (lenValue >= 2) {
+      return null;
+    }
+  }
+
+  for (const [key, value] of Object.entries(robot)) {
+    invertedRobot[value] = key;
+  }
```
witflash, line 33: You can do this task with this cycle only. You can check if `invertedRobot` already has a current key (using `invertedRobot.hasOwnProperty(value)`) and return `null` for this case.

witflash (changes requested):
This solution is not common but it's an interesting approach. 
But you can do it in a simpler way. Just pay attention to my comment and try to improve your code a bit.

PR #2650, opened 2023-09-17
Develop

robsza1122:


src/inverseRobot.js
```diff
@@ -7,7 +7,19 @@
 */
 
 function inverseRobot(robot) {
-  // write code here
+  const fixxedRobot = {};
```
darokrk, line 10: rename this value to be just `fixedRobot` instead of double `xx`

darokrk (approved):
Good job 👏🏽  just fix my last mention and push it, don't need to create another pull request 😃

PR #2661, opened 2023-10-26
Solution

mxpcsr:


src/inverseRobot.js
```diff
@@ -6,8 +6,27 @@
...
+  const invertedRobot = {};
+
+  for (const key in robot) {
+    const value = robot[key];
+
+    if (invertedRobot[value] !== undefined
+
+       || Object.values(invertedRobot).includes(value)) {
+      return null;
+    }
```
nazarmatsevych, line 24:
```suggestion
    if (invertedRobot[value] || Object.values(invertedRobot).includes(value)) {
      return null;
    }
```

Actually the second condition looks redundant

nazarmatsevych (approved):
Good work, left minor comment :)

